PR #16, opened 2020-03-31
fix: pass timeout to 'PartialRowsData.response_iterator'

### LINKED ISSUE #6: Bigtable: read_rows: no deadline causing stuck clients; no way to change it
`Table.read_rows` does not set any deadline, so it can hang forever if the Bigtable server connection hangs. We see this happening once every week or two when running inside GCP, which causes our server to get stuck indefinitely. There appears to be no way in the API to set a deadline, even though [the documentation](https://googleapis.github.io/google-cloud-python/latest/bigtable/table.html#google.cloud.bigtable.table.Table.read_rows) says that the `retry` parameter should do this. Due to a bug, it does not.

### Details:

We are calling `Table.read_rows` to read ~2 rows from BigTable. Using pyflame on a stuck process, both worker threads were waiting on Bigtable, with the stack trace below. I believe the bug is the following:

1. Call `Table.read_rows`. [This calls `PartialRowsData`](https://github.com/googleapis/google-cloud-python/blob/master/bigtable/google/cloud/bigtable/table.py#L436), passing the `retry` argument which defaults to `DEFAULT_RETRY_READ_ROWS`. The default misleadingly sets `deadline=60.0`. ; It also passes `read_method=self._instance._client.table_data_client.transport.read_rows` to `PartialRowsData`, which is a method on `BigtableGrpcTransport`.
2. [`PartialRowsData.__init__` calls `read_method()`](https://github.com/googleapis/google-cloud-python/blob/master/bigtable/google/cloud/bigtable/row_data.py#L398); this is actually raw gRPC `_UnaryStreamMultiCallable`, *not* the gapic `BigtableClient.read_rows`, which AFAICS, is never called. Hence, this gRPC streaming call is started with any deadline. 
3. `PartialRowsData.__iter__` calls `self._read_next_response`, which [calls `return self.retry(self._read_next, on_error=self._on_error)()`](https://github.com/googleapis/google-cloud-python/blob/master/bigtable/google/cloud/bigtable/row_data.py#L454). This gives the impression that `retry` is used, but if I understand gRPC streams correctly, I'm not sure that even makes sense. I think even if the gRPC stream return some error, calling `next` won't actually retry the gRPC, it will just immediately raise the same exception. To retry, I believe you need to actually restart it by calling `read_rows` again.
4. If the Bigtable server now "hangs", the client hangs forever.

### Possible fix:

Change `Table.read_rows` call the gapic `BigtableClient.read_rows` with the `retry` parameter., and change `PartialRowsData.__init__` to take this response iterator, and not take a `retry` parameter at all. This would at least allow setting the gRPC streaming call deadline, although I don't think it will make retrying work (since I *think* the gRPC streaming client will just immediately returns an iterator without actually waiting for a response from the server?)

I haven't actually tried implementing this to see if it works. For now, we will probably just make a raw gRPC read_rows call so we can set an appropriate timeout. 

#### Environment details

*OS*: Linux, ContainerOS (GKE), Container is Debian9 (using distroless)
*Python*: 3.5.3
*API*: google-cloud-bigtable 0.33.0

#### Steps to reproduce

This program loads the Bigtable emulator with 1000 rows, calls `read_rows(retry=DEFAULT.with_deadline(5.0))`, then sends `SIGSTOP` to pause the emulator. This SHOULD cause a `DeadlineExceeded` exception to be raised after 5 seconds. Instead, it hangs forever.

1. Start the Bigtable emulator: `gcloud beta emulators bigtable start`
2. Find the PID: `ps ax | grep cbtemulator`
3. Run the following program with `BIGTABLE_EMULATOR_HOST=localhost:8086 python3 bug.py $PID`

```python
from google.api_core import exceptions
from google.cloud import bigtable
from google.rpc import code_pb2
from google.rpc import status_pb2
import os
import signal
import sys

COLUMN_FAMILY_ID = 'column_family_id'

def main():
    emulator_pid = int(sys.argv[1])
    client = bigtable.Client(project="testing", admin=True)
    instance = client.instance("emulator")

    # create/open a table
    table = instance.table("emulator_table")
    column_family = table.column_family(COLUMN_FAMILY_ID)
    try:
        table.create()
        column_family.create()
    except exceptions.AlreadyExists:
        print('table exists')

    # write a bunch of data
    for i in range(1000):
        k = 'some_key_{:04d}'.format(i)
        print(k)
        row = table.row(k)
        row.set_cell(COLUMN_FAMILY_ID, 'column', 'some_value{:d}'.format(i) * 1000)
        result = table.mutate_rows([row])
        assert len(result) == 1 and result[0].code == code_pb2.OK
        assert table.read_row(k) is not None

    print('starting read')
    rows = table.read_rows(retry=bigtable.table.DEFAULT_RETRY_READ_ROWS.with_deadline(5.0))
    rows_iter = iter(rows)
    r1 = next(rows_iter)
    print('read', r1)
    os.kill(emulator_pid, signal.SIGSTOP)
    print('sent sigstop')
    for r in rows_iter:
        print(r)
    print('done')


if __name__ == '__main__':
    main()
```

#### Stack trace of hung server (using slightly older version of the google-cloud-bigtable library

```
/usr/local/lib/python2.7/threading.py:wait:340
/usr/local/lib/python2.7/site-packages/grpc/_channel.py:_next:348
/usr/local/lib/python2.7/site-packages/grpc/_channel.py:next:366
/usr/local/lib/python2.7/site-packages/google/cloud/bigtable/row_data.py:_read_next:426
/usr/local/lib/python2.7/site-packages/google/api_core/retry.py:retry_target:179
/usr/local/lib/python2.7/site-packages/google/api_core/retry.py:retry_wrapped_func:270
/usr/local/lib/python2.7/site-packages/google/cloud/bigtable/row_data.py:_read_next_response:430
/usr/local/lib/python2.7/site-packages/google/cloud/bigtable/row_data.py:__iter__:441
```

mf2199:
This offers a minimally invasive solution to the client hanging [issue](https://github.com/googleapis/python-bigtable/issues/6). As reported by the author and reproduced using the emulator, the client hangs when there is no response from the CBT server. One of the causes is the specified `deadline` not propagating to the corresponding gRPC calls. The proposed change links this input with the `_grpc._UnaryStreamMultiCallable.__call__.timeout` parameter by adding it as a keyword argument to the `PartialRowsData` constructor.

An alternative way would invoke gapic `wrap_method` to wrap the `bigtable_data_client.transport.read_rows`, along with the user-supplied `timeout` value. But that would bind `Table.read_rows` to gapic retry engine instead of the 'smart' retries offered by 'raw' gRPC. The latter supposedly avoids repeating the whole read sequence in a case of a transient failure, the lack of which may become an issue with large reads, when chances of error become significant while the time needed to complete the sequence gets increasingly large. Another possible approach is to call gapic's `bigtable_data_client.read_rows` instead of `bigtable_data_client.transport.read_rows`. But that would also invoke gapic retries and may cause the same problem as described above.

The `timeout` parameter must be somewhat greater than the value contained in `self.retry`, in order to avoid race-like condition and allow registering the first deadline error before invoking the retry. Otherwise there is a risk of getting stuck in an infinite loop that resets the timeout counter just before it being triggered. The increment by 1 second here is customary but should not be much less than that.

**\*\*\* NOTE: The kokoro tests are bound to fail as the `deadline` property of the `Retry` class has not been exposed yet. This is being addressed in a separate [PR](https://github.com/googleapis/python-api-core/pull/20). \*\*\***

_The coverage test does not register these changes, hence adding a unit test for now has been left as an option._

Fixes #6  🦕


mf2199:
@kolea2 Friendly ping.

google/cloud/bigtable/row_data.py
```diff
@@ -400,7 +400,7 @@ def __init__(self, read_method, request, retry=DEFAULT_RETRY_READ_ROWS):
         self.read_method = read_method
         self.request = request
         self.retry = retry
-        self.response_iterator = read_method(request)
+        self.response_iterator = read_method(request, timeout=self.retry._deadline + 1)
```
kolea2, line 403: is there a test we can add here? (or are we waiting on the release in python-api-core?)

mf2199: @kolea2 The injected `retry.timeout` parameter propagates implicitly to `grpcio.experimental.aio._channel.UnaryStreamMultiCallable` class, hence the best way to test it within the scope of CBT library is via checking the contents of `self.retry` member of the `row_data.PartialRowData` class. I've added such a line to an existing test method along with a couple of other fixes for tests that failed upon introducing the proposed changes.

tests/unit/test_row_data.py
```diff
@@ -385,10 +385,14 @@ def test_constructor(self):
...
     def test_constructor_with_retry(self):
+        from google.cloud.bigtable.row_data import DEFAULT_RETRY_READ_ROWS
+
         client = _Client()
         client._data_stub = mock.MagicMock()
-        request = retry = object()
+        request = object()
+        retry = DEFAULT_RETRY_READ_ROWS
         partial_rows_data = self._make_one(client._data_stub.ReadRows, request, retry)
+        # self.assertEqual(partial_rows_data.response_iterator.timeout, 60.0)
```
tseaver, line 395: Can this line be uncommented now that the `api_core` feature is released?  Better yet, can we mock out the `read_method` and add an assertion that it was called with the expected timeout?

mf2199: @tseaver Absolutely, thanks for pointing this one out. PTAL.

google/cloud/bigtable/row_data.py
kolea2, line 403: why is this + 1? Can you add a comment explaining?

mf2199: @kolea2 Sure, copied from the PR description.

PR #18, opened 2020-04-17
feat(bigtable): skip system tests failing with emulator

### LINKED ISSUE #5: Run system test suite on emulator as well as live service
Some of our libraries have emulator support and it would be a nice to have if the emulator was used to run our system test suites.

I think starting with bigtable or firestore would make sense, then we could move to others if desired.

- Firestore
- Bigtable
- Datastore

HemangChothani:
Fixes #5 

- Skipping the tests which are failing with emulator.

### DIFF
--- a/tests/system.py
+++ b/tests/system.py
@@ -657,6 +657,13 @@ class TestTableAdminAPI(unittest.TestCase):
         for table in self.tables_to_delete:
             table.delete()
 
+    def _maybe_emulator_skip(self, message):
+        # NOTE: This method is necessary because ``Config.IN_EMULATOR``
+        #       is set at runtime rather than import time, which means we
+        #       can't use the @unittest.skipIf decorator.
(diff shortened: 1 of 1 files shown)


HemangChothani:
The 'docs' test failed which is irrelevant to the changes, issue has fixed in PR [20](https://github.com/googleapis/python-bigtable/pull/20)

HemangChothani:
System test failed is not relevant to changes, it throws error like Operation successfully rolled back : Insufficient node quota

tests/system.py
```diff
@@ -657,6 +657,13 @@ def tearDown(self):
         for table in self.tables_to_delete:
             table.delete()
 
+    def _maybe_emulator_skip(self, message):
```
kolea2, line 660: my preference would be to name this `_skip_if_emulated`

PR #20, opened 2020-04-20
docs(bigtable): add parser in extensions of conf.py file

### LINKED ISSUE #19: Building the docs breaks
I noticed that building the docs locally on master fails due to the following error:

sphinx.errors.SphinxWarning: /python-bigtable/docs/index.rst:27:toctree contains reference to document 'changelog' that doesn't have a title: no link will be generated

It turns out that the issue does not occur with the older version of Sphinx (< 3.x), thus it appears as a bug in the Sphinx package itself. A workaround is to pin that dependency to an older version

/cc @crwilcox

HemangChothani:
Fixes #19

Found the relative information here:
https://github.com/readthedocs/recommonmark


HemangChothani:
System test failed is not relevant to changes, it throws error like Operation successfully rolled back : Insufficient node quota

kolea2:
These changes were merged in https://github.com/googleapis/python-bigtable/pull/38

PR #30, opened 2020-05-08
feat: add 'Rowset.add_row_range_with_prefix'

### LINKED ISSUE #24: Reading by prefix
We should add a read by prefix to the Bigtable client. It is a very common use case and can be done like this:
```
    client = bigtable.Client(project=project_id, admin=True)
    instance = client.instance(instance_id)
    table = instance.table(table_id)
    prefix = "phone#"
    end_key = prefix[:-1] + chr(ord(prefix[-1]) + 1)

    row_set = RowSet()
    row_set.add_row_range_from_keys(prefix.encode("utf-8"),
                                    end_key.encode("utf-8"))

    rows = table.read_rows(row_set=row_set)
```

HemangChothani:
Fixes #24


HemangChothani:
System test failed is not relevant to changes, it throws error like `Operation successfully rolled back : Insufficient node quota`

google/cloud/bigtable/row_set.py
```diff
@@ -109,6 +109,24 @@ def add_row_range_from_keys(
         row_range = RowRange(start_key, end_key, start_inclusive, end_inclusive)
         self.row_ranges.append(row_range)
 
+    def add_row_range_by_prefix_from_keys(self, row_key_prefix):
```
crwilcox, line 112: This name seems a bit verbose. Could we name this something a bit more concise?
`add_range_with_prefix` for instance?

docs/snippets_table.py
```diff
@@ -620,7 +620,7 @@ def test_bigtable_create_update_delete_column_family():
     assert column_family_id not in column_families
 
 
-def test_bigtable_add_row_add_row_range_add_row_range_from_keys():
+def test_bigtable_add_row_add_row_range_add_row_range_from_keys_add_row_range_by_prefix_from_keys():
```
kolea2, line 623: can we separate this out into a new test?

tseaver:
[System test failures](https://source.cloud.google.com/results/invocations/e5ff45e3-2c4d-4b7c-85ba-016a10c025b4/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-bigtable%2Fpresubmit%2Fpresubmit/log) due to leaked nodes (exceeding quota) and a (maybe related) 401 in the subsequent test's teardown.

PR #38, opened 2020-05-28
docs: add samples from bigtable

busunkim96:
Samples from https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/bigtable

Samples are unaltered.


busunkim96:
@kolea2 Any idea what's up with the system tests? It seems to consistently be the same failure

PR #47, opened 2020-06-01
docs(bigtable): remove missing argument from instance declaration

### LINKED ISSUE #42: Documentation: Fix outdated Instance declaration
**Issue**
The signature for `client.instance` is inconsistent between the [Using the API docs](https://googleapis.dev/python/bigtable/latest/instance.html#module-google.cloud.bigtable.instance) and the [Instance Admin API docs](https://googleapis.dev/python/bigtable/latest/instance-api.html#instance-factory) 

**Fix**
The [latter is incorrect](https://github.com/googleapis/python-bigtable/blob/master/google/cloud/bigtable/instance.py#L102-L110), and should be amended.

rafilong:
Fixes #42 🦕


rafilong:
Now that location is specified in the cluster, is it worth mentioning this in the documentation about creating an instance or is this something that users can/should discover later?

PR #80, opened 2020-07-31
docs: add sample for writing data with Beam

billyjacobson:


busunkim96:
@billyjacobson Please use `docs: `. `feat` or `fix` will result in a bump in the library version. 😄

samples/beam/requirements.txt
```diff
@@ -0,0 +1,3 @@
+apache-beam==2.23.0
+google-cloud-bigtable==1.4.0
+google-cloud-core==1.3.0
```
busunkim96, line 3: This should be pulled in as a transitive dependency via `google-cloud-bigtable`.

samples/beam/hello_world_write_test.py
```diff
@@ -0,0 +1,57 @@
...
+import os
+import uuid
+
+from google.cloud import bigtable
+import pytest
+
+import hello_world_write
+
+PROJECT = os.environ['GOOGLE_CLOUD_PROJECT']
+BIGTABLE_INSTANCE = os.environ['BIGTABLE_INSTANCE']
```
busunkim96, line 23: Where/how is this environment variable being set in the test environment?

samples/beam/hello_world_write_test.py
```diff
@@ -0,0 +1,57 @@
...
+    yield table_id
+
+    table.delete()
+
+
+def test_hello_world_write(table_id):
+    hello_world_write.run([
+        '--bigtable-project=%s' % PROJECT,
+        '--bigtable-instance=%s' % BIGTABLE_INSTANCE,
+        '--bigtable-table=%s' % table_id])
```
busunkim96, line 47: nit: please use f-strings for new code

```suggestion
        f'--bigtable-project={PROJECT}',
        f'--bigtable-instance={BIGTABLE_INSTANCE}',
        f'--bigtable-table={table_id}'])
```

samples/beam/hello_world_write_test.py
billyjacobson: These are the same variables I've used for other tests, so I assumed they were still available since all the other snippets that were moved are still working. 

They should be defined during [test-samples.sh](https://github.com/googleapis/python-bigtable/blob/master/.kokoro/test-samples.sh#L53) 
They are stored in a secret in the cloud and then [decrypted](https://github.com/googleapis/python-bigtable/blob/master/scripts/decrypt-secrets.sh)

billyjacobson:
@busunkim96 what is "docs-presubmit" It keeps failing for this PR, but I can't figure out what it is testing or how to fix it

busunkim96 (approved):

busunkim96:
@billyjacobson That job isn't set up yet on the GitHub side (it's OK to ignore it)

tseaver:
In order to avoid tripping over quota, I will tag this PR as `automerge` once  PR #93 finishes its own auto-merge.

PR #71, opened 2020-07-24
docs(samples): filter cpu query to get the metrics for the correct resources

busunkim96:
Changes are from https://github.com/GoogleCloudPlatform/python-docs-samples/commit/238722a981405e94dc86e0d644648b88547b4c5e#diff-3525030cd333ea673a78db871ccf0e0e.

PR #81, opened 2020-07-31
docs(samples): filter cpu query to get metrics for the correct resources

busunkim96:
Transferring one commit that was left behind in python-docs-samples. Replaces #71 

---

#4237 BT metricscaler - Filter cpu query to get the metrics for the correct resources [(#4238)](https://github.com/GoogleCloudPlatform/python-docs-samples/issues/4238)

* #4237 Added a filter to the cpu & storage queries to get the metrics for the correct resources


busunkim96:
Hammer merging, main Kokoro test is failing on "quota exceeded".

Issue #78, opened 2020-07-31
Refactor to base classes


rafilong:
Towards #77. Create base classes for shared functionality between async and sync clients. Contains non-networked functions.

daniel-sanche:
I'd say it's safe to close this as out of date

This refactor will occur as part of the upcoming v3 release

PR #84, opened 2020-08-03
refactor: create test_base_cluster

rafilong:
Towards #78
Stacked on #79


tests/unit/test_cluster.py
```diff
@@ -19,48 +19,15 @@
 
 from ._testing import _make_credentials
 
+from tests.unit.test_base_cluster import (
```
kolea2, line 22: what's this change for?

rafilong: An attempt at deduping the various cluster test classes. `_Instance` and `_Client` implementation are shared between testers, and constants are the same regardless of sync/async.

rafilong: From what I saw it seems pretty feasible to declare a global helper class for all testers and import `_Instance` and `_Client` from there, but this seemed like a good middle-ground towards that. Happy to make that change in a separate PR though, or to keep it as-is if there's a reason for the current implementation.

kolea2: nope sounds good to me! Thanks for the explanation

Issue #87, opened 2020-08-03
systests / snippets leaking instances


tseaver:
![Screenshot from 2020-08-03 14-40-09](https://user-images.githubusercontent.com/242750/89215805-43a6bc00-d597-11ea-8705-cf5bc32c2cd8.png)

| Instance ID | Creation Date |
| --- | --- |
| g-c-p-1591650987756 | 2020-06-08t21-16-27 |
| g-c-p-1591651011152 | 2020-06-08t21-16-51 |
| g-c-p-1591720949796 | 2020-06-09t16-42-29 |
| g-c-p-1591721062073 | 2020-06-09t16-44-22 |
| g-c-p-1594318861981 | 2020-07-09t18-21-01 |
| g-c-p-d-1591651011152 | 2020-06-08t21-16-51 |
| g-c-p-d-1594318861981 | 2020-07-09t18-21-01 |
| snippet-tests-1596479760618 | 2020-08-03t18-36-00 |
| snippet-1561580632585 | 2019-06-26t20-23-52 |
| snippet-1596479773184 | 2020-08-03t18-36-13 |
| snippet-tests-1588356586020 | 2020-05-01t18-09-46 |

Of those, two represent what seem to be currently running CI jobs.  I will delete the rest, but we need to find and plug the leaks.

tseaver:
This morning we have [node quota failure](https://source.cloud.google.com/results/invocations/2fda3c2a-7375-4d5d-999d-0a5a555f8f65/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-bigtable%2Fpresubmit%2Fpresubmit/log) due this time entirely to instances leaked from `snippets`.

PR #89, opened 2020-08-03
tests: drop instance / cluster node counts for quota

tseaver:
Toward #87


kolea2:
LGTM - looks like samples tests are failing due to cluster size changes. Will approve once those assertions are fixed.

PR #92, opened 2020-08-04
chore: remove unused snippets

busunkim96:
It looks to me like the sample files in `docs/` are unused. I went through a good number of the region tags and did not find any that were currently published to devsite. Some samples seem to have been moved to other files ([bigtable_create_cluster](https://github.com/googleapis/python-bigtable/search?q=bigtable_create_cluster&unscoped_q=bigtable_create_cluster)) and others seem to be obsolete.

@kolea2 @billyjacobson Could you confirm that it is OK to delete these?


billyjacobson:
The samples in docs aren't used by the cloud documentation, they are used for the API reference documentation https://googleapis.dev/python/bigtable/latest/index.html

Ideally, where there is overlap it could be nice to use the same one to not have as much duplicate code, but I don't really know much about this or what other products do.

busunkim96:
@billyjacobson Ah I see. I'll take a closer look then to see which ones we can pull out.

tseaver:
@busunkim96 Maybe we need to add a `doctest` session to `noxfile.py`:  that way, any snippets that remain in `docs/snippets.py` are tested, and the removing one or more inadvertently would cause a test failure.

PR #100, opened 2020-08-05
chore: add docfx session to noxfile, normalize

tseaver:
Unbreaking the `docs-presubmit` build.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -21,9 +21,12 @@ import shutil
 import nox
 
 
+DEFAULT_PYTHON_VERSION = "3.8"
+SYSTEM_TEST_PYTHON_VERSIONS = ["2.7", "3.8"]
+UNIT_TEST_PYTHON_VERSIONS = ["2.7", "3.5", "3.6", "3.7", "3.8"]
 LOCAL_DEPS = ()
 
-@nox.session(python="3.7")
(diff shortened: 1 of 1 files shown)


tmatsuo:
@tseaver Thanks, I think we should also run synthtool to have other changes, in order to see if the test is actually passing.

tseaver:
@tmatsuo I lost my synthtool-fu during my hiatus.  Can I just get my local copy updated to current `master` and run?  Are there new OS-level dependencies?

tmatsuo:
@tseaver Maybe I can add a commit to the branch.

For the record, I was told there is a dependency on bazel.

tmatsuo:
I think you rebased, that should work too. Thanks!

tseaver:
@tmatsuo I merged #99 to `master`, and then merged it here, which seems to be getting `docs-presubmit` to actually run.

tmatsuo:
The docs-presubmit is passing :)

Do you know why the samples are failing?

tmatsuo:
At a glance, the test has sth like `time.sleep(10)`, so they might be eventually consistent tests.

tseaver:
@tmatsuo We are hitting quota issues, breaking snippets / systests / samples.

PR #128, opened 2020-09-18
feat: add response status to DirectRow.commit()

### LINKED ISSUE #127: DirectRow.commit() should return response state
When I run direct_row.commit(), I thought it succeeded but it didn't. It didn't give me any responses saying the error code or something. It turned out that I was using a wrong column family ID.

The behaviour of DirectRow.commit() should return a response status, same as Table.mutate_rows() because DirectRow.commit() actually calls Table.mutate_rows().

ryanyuan:

Fixes #127  🦕

### DIFF
--- a/tests/unit/test_row.py
+++ b/tests/unit/test_row.py
@@ -359,8 +359,9 @@ class TestDirectRow(unittest.TestCase):
         row.commit()
         self.assertEqual(table.mutated_rows, [row])
 
-    @mock.patch("google.cloud.bigtable.table.Table.mutate_rows")
-    def test_commit_with_exception(self, mock_mutate_rows):
+    def test_commit_with_exception(self):
+        from google.rpc import status_pb2
+
         project_id = "project-id"
         row_key = b"row_key"
         table_name = "projects/more-stuff"
@@ -377,9 +378,9 @@ class TestDirectRow(unittest.TestCase):
 
         # Perform the method and check the result.
         row.set_cell(column_family_id, column, value)
-        from google.rpc import status_pb2
         result = row.commit()
-        self.assertIsInstance(result, status_pb2.Status)
+        expected = status_pb2.Status(code=0)
+        self.assertEqual(result, expected)
 
 
 class TestConditionalRow(unittest.TestCase):
@@ -854,6 +855,7 @@ class _Table(object):
         self.mutated_rows = []
 
     def mutate_rows(self, rows):
-        self.mutated_rows.extend(rows)
         from google.rpc import status_pb2
-        return [status_pb2.Status()]
+
+        self.mutated_rows.extend(rows)
+        return [status_pb2.Status(code=0)]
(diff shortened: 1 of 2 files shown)


tests/unit/test_row.py
```diff
@@ -359,6 +359,28 @@ def test_commit(self):
...
+        client = self._make_client(
+            project=project_id, credentials=credentials, admin=True
+        )
+        table = _Table(table_name, client=client)
+        row = self._make_one(row_key, table)
+        value = b"bytes-value"
+
+        # Perform the method and check the result.
+        row.set_cell(column_family_id, column, value)
+        from google.rpc import status_pb2
```
tseaver, line 380: Please move the import to the top of the testcase method, and add a blank line after (to appease `black`).

tests/unit/test_row.py
```diff
@@ -359,6 +359,28 @@ def test_commit(self):
...
+        result = row.commit()
+        self.assertIsInstance(result, status_pb2.Status)
```
tseaver, line 382: Please test against the expected value from `_Table.mutate_rows` below, e.g.:
```python
        expected = status_pb2.Status(code=0)
        self.assertEqual(result, expected)
```

tests/unit/test_row.py
```diff
@@ -833,3 +855,5 @@ def __init__(self, name, client=None, app_profile_id=None):
 
     def mutate_rows(self, rows):
         self.mutated_rows.extend(rows)
+        from google.rpc import status_pb2
+        return [status_pb2.Status()]
```
tseaver, line 859: Please move the import to the top of the method, and return an actual status code, e.g.:

```python
    def mutate_rows(self, rows):
        from google.rpc import status_pb2
        
        self.mutated_rows.extend(rows)
        return [status_pb2.Status(code=0)]
```

tests/unit/test_row.py
```diff
@@ -359,6 +359,28 @@ def test_commit(self):
         row.commit()
         self.assertEqual(table.mutated_rows, [row])
 
+    @mock.patch("google.cloud.bigtable.table.Table.mutate_rows")
```
tseaver, line 362: This patch is not used (you stub in a `_Table` instance).  Please remove it.

tseaver (changes requested):

tests/unit/test_row.py
ryanyuan: Done.

tests/unit/test_row.py
ryanyuan: Done.

tests/unit/test_row.py
ryanyuan: Done.

tests/unit/test_row.py
ryanyuan: Done.

tseaver (approved):

tseaver:
[Unrelated systest failure](https://source.cloud.google.com/results/invocations/fc1adc69-5fda-4d19-8ff9-b084d943cd9a/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-bigtable%2Fpresubmit%2Fpresubmit/log) reported in #140.

PR #165, opened 2020-11-12
chore: tweak version lookup for non-admin client

tseaver:
See #150

Closes #162
Closes #163

### DIFF
--- a/synth.py
+++ b/synth.py
@@ -45,7 +45,12 @@ library = gapic.py_library(
 s.move(library / "google/cloud/bigtable_admin_v2")
 s.move(library / "tests")
 
-# Work around non-standard installations
+# ----------------------------------------------------------------------------
+# Work around non-standard installations (missing setuptools).
+#
+# These replacements can be removed after migrating to the microgenerator,
+# which will generate them directly.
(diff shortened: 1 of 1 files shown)


synth.py
```diff
@@ -69,6 +69,23 @@
 """
 )
 
+s.replace(
```
kolea2, line 72: is this something that should be templated at the synthtool level instead?

CC @busunkim96

busunkim96: I think the microgenerator templates already behave like this. They don't raise if a version cannot be found. So this can probably be removed when the repo is transitioned. https://github.com/googleapis/gapic-generator-python/blob/af59c2c3c3d6b7e1f626c3fbc2c03f99ca31b4a4/gapic/templates/%25namespace/%25name_%25version/%25sub/services/%25service/client.py.j2#L715-L722

kolea2: ok, since that is still in progress - @tseaver mind adding a comment this can be removed later?

tseaver: @kolea2 Done in 2d703ef

kolea2 (approved):

PR #178, opened 2020-12-02
feat: support filtering on incrementable values

### LINKED ISSUE #177: Add support for filtering based on incrementable values
/cc @billyjacobson 

Currently, the Python bigtable library support for filtering on numeric values expects raw bytes, e.g.:

```python
from google.cloud.bigtable import row_filters

filter = row_filters.ValueRangeFilter(
    start_value=b'\x00\x00\x00\x00\x00\x00\x00\x01',  # 1
    end_value=b'\x00\x00\x00\x00\x00\x00\x00\x0a',  # 10
)
```

We already have support for *writing* "normal" integers, e.g.:

```python
from google.cloud.bigtable import row_data

row = row_data.DirectRow(b'row-key')
row.set_cell('col-family', b'column', value=1)
```
where the integer `1` gets converted to `b'\x00\x00\x00\x00\x00\x00\x00\x01'` under the covers.

This feature request is to add similar support for filtering on values, e.g.:

```python
from google.cloud.bigtable import row_filters

filter = row_filters.ValueRangeFilter(start_value=1, end_value=10)
```

While we are at it, add an `ExactValueFilter` wrapper around `` (for parity with other languages), e.g..

```python
from google.cloud.bigtable import row_filters

filter = row_filters.ExactValueFilter(value=1)
```

which would map to (for integer values):

```python
filter = row_filters.ValueRegexFilter(b'\x00\x00\x00\x00\x00\x00\x00\x01')
```

tseaver:
Closes #177.

### DIFF
--- a/google/cloud/bigtable/row_filters.py
+++ b/google/cloud/bigtable/row_filters.py
@@ -14,10 +14,14 @@
 
 """Filters for Google Cloud Bigtable Row classes."""
 
+import struct
+
 
 from google.cloud._helpers import _microseconds_from_datetime
 from google.cloud._helpers import _to_bytes
 from google.cloud.bigtable_v2.proto import data_pb2 as data_v2_pb2
+
+_PACK_I64 = struct.Struct(">q").pack
 
 
 class RowFilter(object):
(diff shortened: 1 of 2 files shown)


tseaver:
@billyjacobson PTAL.

billyjacobson (approved):
Looks good from a Bigtable perspective.

PR #182, opened 2020-12-04
feat: add keep alive timeout

### LINKED ISSUE #181: Add keepalive in grpc channel for Bigtable client
Sets KeepAlive timer to 30s (Google Frontends are configured limits keepalive calls at 30s by default. If used below 30s, grpc automatically increases keepalive time by 2x after too_many_ping commands).

Sets KeepAliveTimeout to 10s (conservative)

Sets keepAliveWithoutCalls to false as it can potentially increase load.

Reference issue: https://github.com/googleapis/java-bigtable-hbase/issues/2681

HemangChothani:
Fixes #181 

Added  `grpc.keepalive_time_ms` and `grpc.keepalive_timeout_ms` for an emulator as it used insecure channel.

Clients are gapic generated client so need to add in respective files:

1: https://github.com/googleapis/python-bigtable/blob/9b380a36df0652511639da7c128b80d39db1bc0c/google/cloud/bigtable_v2/gapic/transports/bigtable_grpc_transport.py#L67-L74

2 : https://github.com/googleapis/python-bigtable/blob/9b380a36df0652511639da7c128b80d39db1bc0c/google/cloud/bigtable_admin_v2/gapic/transports/bigtable_instance_admin_grpc_transport.py#L71-L79

3: https://github.com/googleapis/python-bigtable/blob/9b380a36df0652511639da7c128b80d39db1bc0c/google/cloud/bigtable_admin_v2/gapic/transports/bigtable_table_admin_grpc_transport.py#L71-L79

### DIFF
--- a/google/cloud/bigtable/client.py
+++ b/google/cloud/bigtable/client.py
@@ -35,6 +35,11 @@ from google.api_core.gapic_v1 import client_info
 
 from google.cloud import bigtable_v2
 from google.cloud import bigtable_admin_v2
+from google.cloud.bigtable_v2.gapic.transports import bigtable_grpc_transport
+from google.cloud.bigtable_admin_v2.gapic.transports import (
+    bigtable_table_admin_grpc_transport,
+    bigtable_instance_admin_grpc_transport,
+)
 
 from google.cloud.bigtable import __version__
 from google.cloud.bigtable.instance import Instance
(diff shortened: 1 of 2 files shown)


kolea2:
CC @sushanb

google/cloud/bigtable/client.py
```diff
@@ -161,7 +161,13 @@ def __init__(
         self._emulator_channel = None
 
         if self._emulator_host is not None:
-            self._emulator_channel = grpc.insecure_channel(self._emulator_host)
+            self._emulator_channel = grpc.insecure_channel(
```
kolea2, line 164: is there another place this should be added for non emulator settings?

kolea2: ignore, seeing your comments now

kolea2:
Is there a way to override the channel settings in the handwritten layer?

HemangChothani:
Yes, I think so. We need to create channel explicitly in handwritten client for all the classes which mentioned above and then create a respective transport class object and  pass as an argument for all the classes,so gapic class doesn't create or override the transport class object.

crwilcox (approved):

PR #186, opened 2020-12-27
feat: backup restore to another Instance

mf2199:
**The summary of proposed changes:**

- A new optional argument added to `Backup.restore` method, allowing users to specify Instance to restore the backup into, if different from the one the backup was created in.

**ToDo:**

- (edit: kolea2) update region to `LOCATION_ID` global variable.

### DIFF
--- a/tests/system.py
+++ b/tests/system.py
@@ -876,7 +876,9 @@ class TestTableAdminAPI(unittest.TestCase):
 
         # Testing `Backup.update_expire_time()` method
         expire += 3600  # A one-hour change in the `expire_time` parameter
-        temp_backup.update_expire_time(datetime.datetime.utcfromtimestamp(expire))
+        temp_backup.update_expire_time(
+            datetime.datetime.utcfromtimestamp(expire),
+        )
 
         # Testing `Backup.get()` method
         temp_table_backup = temp_backup.get()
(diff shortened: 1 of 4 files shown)


tests/unit/test_backup.py
```diff
@@ -724,15 +729,22 @@ def test_restore_success(self):
             expire_time=timestamp,
         )
 
-        future = backup.restore(self.TABLE_ID)
+        future = backup.restore(self.TABLE_ID, instance_id)
```
kolea2, line 732: can we make this a new test instead of changing the existing?

kolea2 (changes requested):
Looking good, please note I switched the target branch from master to `backups-different-instance`

tests/system.py
```diff
@@ -886,12 +888,37 @@ def test_backup(self):
...
+        alt_instance = Config.CLIENT.instance(alt_instance_id, labels=LABELS)
+        alt_cluster = alt_instance.cluster(
+            cluster_id=alt_cluster_id,
+            location_id="us-east1-c",  # TODO: Change to default `LOCATION_ID`
+            serve_nodes=SERVE_NODES,
+        )
+        if not Config.IN_EMULATOR:
+            alt_instance.create(clusters=[alt_cluster]).result(timeout=10)
+
+        # Testing `reastore()`...
```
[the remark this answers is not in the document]
mf2199: 👍 Thanks! Done.

tests/unit/test_backup.py
mf2199: That's basically the idea. In order to avoid code repetition I created this common helper to be called by two separate tests (L743-L747):
```
    def test_restore_default(self):
        self._restore_helper()

    def test_restore_to_another_instance(self):
        self._restore_helper(self.ALT_INSTANCE, self.ALT_INSTANCE_NAME)
```
Would you still prefer having two fully coded, nearly identical tests instead?

kolea2: I would like one test that calls `restore` as is, and one test that calls restore passing in an instance id. If those are both covered, I am good 👍

kolea2 (approved):
This is looking good, merging into `backup-different-instance` branch as a WIP

PR #203, opened 2021-02-04
feat!: microgenerator changes

kolea2:
Release-As: v2.0.0-dev1

Regen the client with the microgenerator. This requires the client to be used with Python >= 3.6

There are a few outstanding questions I have that I will annotate in the PR. Most important changes to pay attention to are the handwritten client and test layers.


tests/unit/test_app_profile.py
```diff
@@ -510,49 +531,59 @@ def test_update_app_profile_routing_any(self):
...
-        actual_request = channel.requests[0][1]
+        actual_request = client._instance_admin_client.update_app_profile.call_args_list[
+            0
+        ].kwargs
 
         self.assertEqual(actual_request, expected_request)
-        self.assertIsInstance(result, operation.Operation)
-        self.assertEqual(result.operation.name, self.OP_NAME)
-        self.assertIsInstance(result.metadata, messages_v2_pb2.UpdateAppProfileMetadata)
+        # self.assertIsInstance(result, operation.Operation)
```
kolea2, line 572: @crwilcox I have a few instances like this where after we refactored this test (from using channels), we are no longer getting an `operation.Operation` back, but a pb2 Operation. Are we missing anything by removing these assertions?

tests/unit/test_client.py
```diff
@@ -213,7 +213,8 @@ def test_table_data_client_not_initialized(self):
 
         table_data_client = client.table_data_client
         self.assertIsInstance(table_data_client, BigtableClient)
-        self.assertIs(table_data_client._client_info, _CLIENT_INFO)
+        # todo is this expected?
+        self.assertIs(client._client_info, _CLIENT_INFO)
```
kolea2, line 217: `table_data_client` no longer has `._client_info`. Is this expected?

google/cloud/bigtable/table.py
```diff
@@ -1071,13 +1087,14 @@ def _do_mutate_retryable_rows(self):
...
         try:
             responses = data_client.mutate_rows(
-                self.table_name,
-                entries,
+                table_name=self.table_name,
+                entries=entries,
                 app_profile_id=self.app_profile_id,
                 retry=None,
-                **kwargs
+                # **kwargs
```
kolea2, line 1097: not sure what do do with the `**kwargs` here

tests/unit/test_table.py
```diff
@@ -840,14 +858,33 @@ def test_read_retry_rows(self):
...
+        data_api.table_path.return_value = (
+            "projects/"
+            + self.PROJECT_ID
+            + "/instances/"
+            + self.INSTANCE_ID
+            + "/tables/"
+            + self.TABLE_ID
+        )
+        table_api.table_path.return_value = (
+            "projects/"
```
crwilcox, line 870: these would likely look much nicer as f-strings

tests/unit/test_table.py
```diff
@@ -899,13 +938,29 @@ def test_yield_retry_rows(self):
...
         # Patch the stub used by the API method.
-        client._table_data_client.transport.read_rows = mock.Mock(
-            side_effect=[
-                response_failure_iterator_1,
-                response_failure_iterator_2,
-                response_iterator,
-            ]
-        )
+        data_api.table_path.return_value = (
+            "projects/"
```
crwilcox, line 942: f-string or format string?

tests/unit/test_table.py
```diff
@@ -972,9 +1029,25 @@ def test_yield_rows_with_row_set(self):
         response_iterator = _MockReadRowsIterator(response_1, response_2, response_3)
 
         # Patch the stub used by the API method.
-        client._table_data_client.transport.read_rows = mock.Mock(
-            side_effect=[response_iterator]
-        )
+        data_api.table_path.return_value = (
+            "projects/"
```
crwilcox, line 1033: ditto

tests/unit/test_table.py
```diff
@@ -1313,19 +1387,20 @@ def _list_backups_helper(self, cluster_id=None, filter_=None, **kwargs):
...
+        table._instance._client._table_admin_client = table_api
 
         parent = self.INSTANCE_NAME + "/clusters/cluster"
-        backups_pb = bigtable_table_admin_pb2.ListBackupsResponse(
+        backups_pb = bigtable_table_admin.ListBackupsResponse(
             backups=[
-                table_pb2.Backup(name=parent + "/backups/op1"),
-                table_pb2.Backup(name=parent + "/backups/op2"),
-                table_pb2.Backup(name=parent + "/backups/op3"),
+                backup_pb(name=parent + "/backups/op1"),
```
crwilcox, line 1396: nit: also could be fstrs.

tests/unit/test_table.py
```diff
@@ -1455,12 +1540,15 @@ def _make_responses(self, codes):
...
+            client as bigtable_table_admin,
+        )
 
-        data_api = mock.create_autospec(bigtable_client.BigtableClient)
-        table_api = mock.create_autospec(
-            bigtable_table_admin_client.BigtableTableAdminClient
+        data_api = mock.create_autospec(BigtableClient)
+        table_api = mock.create_autospec(bigtable_table_admin.BigtableTableAdminClient)
+        table_api.table_path.return_value = (
+            "projects/self.PROJECT_ID/instances/self.INSTANCE_ID/tables/self.TABLE_ID"
```
crwilcox, line 1551: is the project name literally `self.PROJECT_ID`?

tests/unit/test_table.py
```diff
@@ -1509,12 +1600,32 @@ def test_callable_no_retry_strategy(self):
...
         row_3.set_cell("cf", b"col", b"value3")
 
-        worker = self._make_worker(client, table.name, [row_1, row_2, row_3])
-
         response_codes = [self.SUCCESS, self.RETRYABLE_1, self.NON_RETRYABLE]
         response = self._make_responses(response_codes)
         data_api.mutate_rows = mock.MagicMock(return_value=[response])
 
+        data_api.table_path.return_value = (
+            "projects/"
```
crwilcox, line 1608: format string?

tests/unit/test_table.py
```diff
@@ -1539,8 +1652,24 @@ def test_callable_retry(self):
         #   - State of responses_statuses should be
         #       [success, success, non-retryable]
 
-        data_api = bigtable_client.BigtableClient(mock.Mock())
-        table_api = bigtable_table_admin_client.BigtableTableAdminClient(mock.Mock())
+        data_api = mock.create_autospec(BigtableClient)
+        table_api = mock.create_autospec(bigtable_table_admin.BigtableTableAdminClient)
+        data_api.table_path.return_value = (
+            "projects/"
+            + self.PROJECT_ID
```
crwilcox, line 1659: f-str

tests/unit/test_table.py
```diff
@@ -1610,8 +1738,24 @@ def test_do_mutate_retryable_rows(self):
         # Expectation:
         #   - Expect [success, non-retryable]
 
-        data_api = bigtable_client.BigtableClient(mock.Mock())
-        table_api = bigtable_table_admin_client.BigtableTableAdminClient(mock.Mock())
+        data_api = mock.create_autospec(BigtableClient)
+        table_api = mock.create_autospec(bigtable_table_admin.BigtableTableAdminClient)
+        data_api.table_path.return_value = (
+            "projects/"
```
crwilcox, line 1744: f-str

tests/unit/test_table.py
```diff
@@ -1678,8 +1825,25 @@ def test_do_mutate_retryable_rows_retry(self):
         )
 
         # Patch the stub used by the API method.
-        inner_api_calls = client._table_data_client._inner_api_calls
-        inner_api_calls["mutate_rows"] = mock.Mock(side_effect=[[response]])
+        client._table_data_client.mutate_rows.side_effect = [[response]]
+        data_api.table_path.return_value = (
+            "projects/"
+            + self.PROJECT_ID
```
crwilcox, line 1831: fstr

tests/unit/test_table.py
```diff
@@ -1735,8 +1901,25 @@ def test_do_mutate_retryable_rows_second_retry(self):
         response = self._make_responses([self.SUCCESS, self.RETRYABLE_1])
 
         # Patch the stub used by the API method.
-        inner_api_calls = client._table_data_client._inner_api_calls
-        inner_api_calls["mutate_rows"] = mock.Mock(side_effect=[[response]])
+        client._table_data_client.mutate_rows.side_effect = [[response]]
+        data_api.table_path.return_value = (
+            "projects/"
```
crwilcox, line 1906: f-str. Also, as much as this is repeated, could we make this a helper?

tests/unit/test_table.py
```diff
@@ -1795,8 +1980,25 @@ def test_do_mutate_retryable_rows_second_try(self):
         response = self._make_responses([self.NON_RETRYABLE, self.SUCCESS])
 
         # Patch the stub used by the API method.
-        inner_api_calls = client._table_data_client._inner_api_calls
-        inner_api_calls["mutate_rows"] = mock.Mock(side_effect=[[response]])
+        client._table_data_client.mutate_rows.side_effect = [[response]]
+        data_api.table_path.return_value = (
+            "projects/"
```
crwilcox, line 1985: .

tests/unit/test_table.py
```diff
@@ -1847,6 +2049,15 @@ def test_do_mutate_retryable_rows_second_try_no_retryable(self):
         worker.responses_statuses = self._make_responses_statuses(
             [self.SUCCESS, self.NON_RETRYABLE]
         )
+        table_api.table_path.return_value = (
+            "projects/"
```
crwilcox, line 2053: .

tests/unit/test_table.py
```diff
@@ -1879,8 +2092,25 @@ def test_do_mutate_retryable_rows_mismatch_num_responses(self):
         response = self._make_responses([self.SUCCESS])
 
         # Patch the stub used by the API method.
-        inner_api_calls = client._table_data_client._inner_api_calls
-        inner_api_calls["mutate_rows"] = mock.Mock(side_effect=[[response]])
+        client._table_data_client.mutate_rows.side_effect = [[response]]
+        data_api.table_path.return_value = (
+            "projects/"
```
crwilcox, line 2097: .

crwilcox (changes requested):
This is so exciting. Thank you for all of your work to accomplish this.

cc: @busunkim96 @danoscarmike

google/cloud/bigtable/table.py
kolea2: leaving them in gives errors in the system tests (IIRC something related to too many arguments passed in)

google/cloud/bigtable/client.py
```diff
@@ -219,7 +217,7 @@ def project_path(self):
         :rtype: str
         :returns: Return a fully-qualified project string.
         """
-        return self.instance_admin_client.project_path(self.project)
+        return "projects/{project}".format(project=self.project)
```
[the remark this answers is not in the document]
busunkim96: I think we re-added `project_path` and `common_project_path`, so calling out to that is another option.

google/cloud/bigtable/table.py
crwilcox: This likely will need investigating then. I can help if needed. But this would likely break something for users.

kolea2: Nevermind, just tried again and it worked :)

docs/snippets.py
```diff
@@ -723,7 +723,7 @@ def test_bigtable_instance_from_pb():
 def test_bigtable_cluster_from_pb():
     # [START bigtable_api_cluster_from_pb]
     from google.cloud.bigtable import Client
-    from google.cloud.bigtable_admin_v2.types import instance_pb2
+    from google.cloud.bigtable_admin_v2.types import instance
```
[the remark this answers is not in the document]
kolea2: This was semi bad search and replace, but the import type still changes. This is in the docs snippets so I'm not sure how used this is? but yes, can add it to the commit message as a highlight and we can add it to the changelog when we release.

google/cloud/bigtable/client.py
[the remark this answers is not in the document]
kolea2: thank you! will revert this back to `project_path`.

google/cloud/bigtable_v2/__init__.py
```diff
@@ -1,42 +1,71 @@
...
-        "A future version of this library will drop support for Python 2.7. "
-        "More details about Python 2 support for Google Cloud Client Libraries "
-        "can be found at https://cloud.google.com/python/docs/python2-sunset/"
-    )
-    warnings.warn(message, DeprecationWarning)
-
-
-class BigtableClient(bigtable_client.BigtableClient):
-    __doc__ = bigtable_client.BigtableClient.__doc__
+from .services.bigtable import BigtableClient
```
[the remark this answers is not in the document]
kolea2: what is the change here as this is generated right?

tests/system.py
```diff
@@ -41,9 +42,10 @@
 from google.cloud.bigtable.row_data import PartialRowData
 from google.cloud.bigtable.row_set import RowSet
 from google.cloud.bigtable.row_set import RowRange
-from google.cloud.bigtable_admin_v2.gapic import (
-    bigtable_table_admin_client_config as table_admin_config,
-)
+
+# from google.cloud.bigtable_admin_v2.gapic import (
```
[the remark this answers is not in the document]
kolea2: Thanks! Forgot to call this out in my comments. It looks like this was used for some tweaking in the next few lines that I didn't know how to replicate with the new microgen code. WDYT?

tests/unit/test_table.py
kolea2: Hahaha, this is such a good catch. This was unnecessary - deleted :)

tests/unit/test_table.py
kolea2: A lot of these were unnecessary leftover from when I was figuring out what to properly mock. Cleaning up.

tests/unit/test_table.py
```diff
@@ -858,22 +858,7 @@ def test_read_retry_rows(self):
...
         # Patch the stub used by the API method.
-        data_api.table_path.return_value = (
-            "projects/"
-            + self.PROJECT_ID
-            + "/instances/"
-            + self.INSTANCE_ID
-            + "/tables/"
-            + self.TABLE_ID
-        )
-        table_api.table_path.return_value = (
```
crwilcox: did you mean to drop the table_api assignment here?

kolea2: yes - some of these assignments were from when I was trying out which items to mock to get the tests to work - this one was unnecessary.

crwilcox (approved):

google/cloud/bigtable/client.py
```diff
@@ -212,23 +212,19 @@ def _create_gapic_client_channel(self, client_class, grpc_transport):
...
                 "grpc.max_send_message_length": -1,
                 "grpc.max_receive_message_length": -1,
                 "grpc.keepalive_time_ms": 30000,
                 "grpc.keepalive_timeout_ms": 10000,
             }.items(),
         )
-        transport = grpc_transport(
-            address=api_endpoint,
-            channel=channel,
-            credentials=None,
```
kolea2, line 230: @crwilcox confirming removing `credentials` here is ok? (looked at firestore)

google/cloud/bigtable/table.py
```diff
@@ -1071,10 +1087,11 @@ def _do_mutate_retryable_rows(self):
         if self.timeout is not None:
             kwargs["timeout"] = timeout.ExponentialTimeout(deadline=self.timeout)
 
+        # todo confirm this change
```
kolea2, line 1090: @crwilcox another todo I missed - I think my concern here was the mutate rows call wasn't wrapped in a `request={...` but I think due to the code below this (will need to expand) this works as expected. LMK if you have a concern

tests/system.py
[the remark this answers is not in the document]
kolea2: @crwilcox ping

tests/unit/test_table.py
```diff
@@ -1340,16 +1400,21 @@ def _list_backups_helper(self, cluster_id=None, filter_=None, **kwargs):
...
+            order_by = kwargs["order_by"]
+
+        if "page_size" in kwargs:
+            page_size = kwargs["page_size"]
+
         api.assert_called_once_with(
-            bigtable_table_admin_pb2.ListBackupsRequest(
-                parent=parent, filter=backups_filter, **kwargs
-            ),
-            retry=mock.ANY,
```
kolea2, line 1350: @crwilcox one last thing to point out - the last few params were now removed from the assertion because we are no longer using the `inner_api_calls` to get the result, let me know if oyu think this should still be here.

google/cloud/bigtable/client.py
crwilcox: should be the same. :)

samples/beam/requirements.txt
```diff
@@ -1,3 +1,3 @@
 apache-beam==2.27.0
-google-cloud-bigtable==1.6.1
+google-cloud-bigtable==1.0.0
```
crwilcox, line 2: this is an odd change. does 1.6.1 not work anymore? Do we need to range specifier (>= 1.6.1 < 2.0.0dev)?

crwilcox (approved):

PR #249, opened 2021-03-17
feat: customer managed keys (CMEK)

### LINKED ISSUE #272: Add CMEK support to Bigtable
Feature checklist:

- [x] Create `EncryptionInfo` class
- [x] Create `Status` class for `EncryptionInfo.encryption_status` field
- [x] Add `encryption_info` field to Backup
- [x] Add `get_encryption_info` method to `Table`
- [x] Add `kms_key_name` field to Cluster

See also the [Java reference implementation](https://github.com/googleapis/java-bigtable/pull/656).

chrisrossi:
Implement customer managed keys (CMEK) feature.

Closes #272


chrisrossi:
System tests only work with a particular project and a particular key. I'll need to get it creating its own key ring and key to use.

CMEK system tests are also, literally, just a copy/paste of the existing system tests with the `kms_key_name` argument added at cluster creation time. There's probably a better way that doesn't replicate so much code. But maybe that's fine.

I need to flesh out a wrapper class, still, for the `Status` PB that is the value for `EncryptionInfo.encryption_status`. ATM, it just returns the raw PB.

google/cloud/bigtable/cluster.py
```diff
@@ -145,6 +160,8 @@ def _update_from_pb(self, cluster_pb):
         self.location_id = cluster_pb.location.split("/")[-1]
         self.serve_nodes = cluster_pb.serve_nodes
         self.default_storage_type = cluster_pb.default_storage_type
+        if cluster_pb.encryption_config:
```
kolea2, line 163: what does this default to if this condition is not met?

google/cloud/bigtable/encryption.py
```diff
@@ -0,0 +1,64 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Class for encryption info for tables and backups."""
+
+from google.cloud.bigtable.error import Status
+
+
+class EncryptionInfo:
```
kolea2, line 20: for clarity, I would rename this file to `encryption_info.py`

google/cloud/bigtable/error.py
```diff
@@ -0,0 +1,64 @@
...
+
+"""Class for error status."""
+
+
+class Status:
+    """A status, comprising a code and a message.
+
+    See: `Cloud APIs Errors <https://cloud.google.com/apis/design/errors>`_
+
+    This is a thin wrapper for ``google.rpc.status_pb2.Status``.
```
kolea2, line 23: is it possible to use the pb2 Status as is? Would be nice to have less handwritten code to look after :)

google/cloud/bigtable/cluster.py
```diff
@@ -356,4 +378,8 @@ def _to_pb(self):
             serve_nodes=self.serve_nodes,
             default_storage_type=self.default_storage_type,
         )
+        if self._kms_key_name:
```
kolea2, line 381: same question here, what does this look like when kms key is not set? Is there a default view?

google/cloud/bigtable/cluster.py
chrisrossi: Oops. If it's being called from the constructor, it will be None, but we can't assume that can we? I'll add an `else` here.

google/cloud/bigtable/error.py
chrisrossi: I was doing it in the first draft, then decided for consistency I should probably wrap it. I think I saw in the Java reference implementation it was wrapped and figured I'd better follow suit. I think my personal preference would be to return the bare pb2, if that's not considered a violation of the contract for user exposed APIs.

chrisrossi: It also just occurred to me, that the wrapper class might be preferable for documentation purposes, as it causes the interface to get documented in the Sphinx docs. Otherwise, the user has to track down the protocol buffer definitions. 

I'll defer to whatever call you want to make here.

google/cloud/bigtable/cluster.py
chrisrossi: When I look at the PBs returned from the service, I'm finding that it's omitting `encryption_config` altogether if using Google default encryption, so this creates a PB the same as we get them from the service.

google/cloud/bigtable/encryption.py
chrisrossi: 090080e

google/cloud/bigtable/cluster.py
chrisrossi: f1aae9d

google/cloud/bigtable/error.py
kolea2: Sorry for the delay here - following up on this. I'll update here when I can :)

google/cloud/bigtable/cluster.py
kolea2: oh sorry missed that, makes sense thanks!

google/cloud/bigtable/error.py
chrisrossi: Oh no, I'm sorry for the delay in getting back to you. I had fumbled something in my email filters and didn't see you had responded to the initial draft, so I thought I was still waiting.

google/cloud/bigtable/cluster.py
chrisrossi: Actually, you correctly pointed out a logical error, which I corrected. Thank you!

google/cloud/bigtable/error.py
kolea2: I think we're going to keep this - resolving

google/cloud/bigtable/error.py
```diff
@@ -0,0 +1,64 @@
...
+    def __init__(self, status_pb):
+        self.status_pb = status_pb
+
+    @property
+    def code(self):
+        """The status code.
+
+        Values are defined in ``google.rpc.code_pb2.Code``.
+
+        See: `googe.rpc.Code
```
kolea2, line 38: nit: typo "google"

tests/unit/test_table.py
```diff
@@ -534,6 +534,87 @@ def test_get_cluster_states(self):
...
+            project="project-id", credentials=credentials, admin=True
+        )
+        instance = client.instance(instance_id=self.INSTANCE_ID)
+        table = self._make_one(self.TABLE_ID, instance)
+
+        response_pb = _TablePB(
+            cluster_states={
+                "cluster-id1": _ClusterStateEncryptionInfoPB(
+                    encryption_type=ENCRYPTION_TYPE_UNSPECIFIED,
+                    encryption_status=_StatusPB(Code.OK, "beats me"),
```
kolea2, line 566: 😄 , we may want to make these a little more realistic - something like "status ok"

kolea2 (commented):
Looking good with two small comments. Will also ask @crwilcox to take a quick look

google/cloud/bigtable/error.py
chrisrossi: e7a45ee

tests/unit/test_table.py
chrisrossi: e7a45ee

kolea2 (approved):
lgtm, will merge after @crwilcox takes a look. Thank you!

crwilcox (approved):

PR #284, opened 2021-04-06
chore(deps): update samples to bigtable 2.0.0

kolea2:
Opening in favor of #283

### DIFF
--- a/samples/beam/requirements.txt
+++ b/samples/beam/requirements.txt
@@ -1,3 +1,3 @@
 apache-beam==2.28.0
-google-cloud-bigtable<2.0.0dev1
+google-cloud-bigtable<2.0.0
 google-cloud-core==1.6.0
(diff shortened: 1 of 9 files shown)


samples/beam/requirements.txt
```diff
@@ -1,3 +1,3 @@
 apache-beam==2.28.0
-google-cloud-bigtable<2.0.0dev1
+google-cloud-bigtable<2.0.0
```
crwilcox, line 2: I think this will include the dev candidate in the allowed versions. This might be unintended?

crwilcox (approved):

samples/beam/requirements.txt
kolea2: I think it's ok - if it were picking up the dev candidate the build would fail, like in #283

PR #290, opened 2021-04-14
test: add tests for instanceadmin sample

### LINKED ISSUE #201: Add tests for instanceadmin samples
@billyjacobson FYI

chrisrossi:
Closes #201


samples/instanceadmin/instanceadmin.py
```diff
@@ -1,259 +1,270 @@
-#!/usr/bin/env python
```
chrisrossi, line 1: I ran `black` on this file. I'm not sure what change it made that makes the entire file different, but that's what happened. I could unblacken, but there are some manual edits I did before running `black` that I'd have to reproduce.

samples/instanceadmin/instanceadmin.py
```diff
@@ -1,259 +1,270 @@
...
+    # [END bigtable_get_instance]
+
+    # [START bigtable_get_clusters]
+    print("\nListing clusters...")
+    for cluster in instance.list_clusters()[0]:
+        print(cluster.cluster_id)
+    # [END bigtable_get_clusters]
+
+
+def create_dev_instance(project_id, instance_id, cluster_id):
```
billyjacobson, line 101: We don't need this anymore. From the docs:

"Note: You are no longer able to create new development instances, because production instances of only one or two clusters are now generally available."

samples/instanceadmin/instanceadmin.py
```diff
@@ -1,259 +1,270 @@
...
+    :param project_id: Project id of the client.
+
+    :type instance_id: str
+    :param instance_id: Instance of the client.
+    """
+    client = bigtable.Client(project=project_id, admin=True)
+    location_id = "us-central1-f"
+    serve_nodes = 1
+    storage_type = enums.StorageType.SSD
+    production = enums.Instance.Type.PRODUCTION
```
billyjacobson, line 55: I'm not sure if the instance_type param is still needed since all are production

samples/instanceadmin/instanceadmin.py
```diff
@@ -1,259 +1,270 @@
...
+    if cluster.exists():
+        cluster.delete()
+        print("Cluster deleted: {}".format(cluster_id))
+    else:
+        print("\nCluster {} does not exist.".format(cluster_id))
+
+    # [END bigtable_delete_cluster]
+
+
+if __name__ == "__main__":
```
billyjacobson, line 231: These samples are very old so they still have this CLI runner. Which might be good to remove and bring these into the newer format?

I think because this is a sample vs snippets they can stay all in one file, but this is something I need to think about from a Bigtable perspective rather than just for this PR

samples/instanceadmin/test_instanceadmin.py
```diff
@@ -0,0 +1,142 @@
...
+        assert f"Instance {INSTANCE} already exists." in out
+        assert "Listing instances" in out
+        assert f"\n{INSTANCE}\n" in out
+        assert f"Name of instance: {INSTANCE}" in out
+        assert "Labels: {'prod-label': 'prod-label'}" in out
+        assert "Listing clusters..." in out
+        assert f"\n{CLUSTER1}\n" in out
+
+    finally:
+        instanceadmin.delete_instance(PROJECT, INSTANCE)
```
billyjacobson, line 55: is there a way to try again if it fails or just make sure it gets deleted? If too many tests fail the deletion the project will run out of Bigtable nodes and fail and they will need to be cleaned up manually. Maybe a check at start of test for instances starting with instanceadmin and deleting? not sure the best approach

samples/instanceadmin/test_instanceadmin.py
```diff
@@ -0,0 +1,142 @@
...
+        out = capsys.readouterr().out
+        assert "Deleting instance" in out
+        assert f"Deleted instance: {INSTANCE}" in out
+
+    finally:
+        instanceadmin.delete_instance(PROJECT, INSTANCE)
+
+
+def test_create_dev_instance(capsys):
```
billyjacobson, line 80: could probably remove (see above note)

samples/instanceadmin/test_instanceadmin.py
```diff
@@ -0,0 +1,142 @@
...
+        out = capsys.readouterr().out
+        assert f"Instance {INSTANCE} already exists" in out
+
+    finally:
+        instanceadmin.delete_instance(PROJECT, INSTANCE)
+
+
+def test_add_and_delete_cluster(capsys):
+    try:
+        # This won't work, because the instance isn't created yet
```
billyjacobson, line 97: this is good, thank you!

billyjacobson (commented):
Looks good overall, just a few things to potentially change. Thanks for taking the time to do this

samples/instanceadmin/instanceadmin.py
chrisrossi: It might make sense to make a separate issue with its own PR for this, since it's wider in scope than just this one test.

samples/instanceadmin/test_instanceadmin.py
chrisrossi: a8f33ac

samples/instanceadmin/instanceadmin.py
chrisrossi: a8f33ac

samples/instanceadmin/instanceadmin.py
chrisrossi: a8f33ac

samples/instanceadmin/test_instanceadmin.py
chrisrossi: 8b30d5c

The trade off is now we can screw up another test run or if there is more than one test run happening in parallel for the same project. I don't really see that coming up in practice, though.

chrisrossi:
Hi @billyjacobson ,

Can you take a look, again, with the updates I've made?

Thanks!

samples/instanceadmin/instanceadmin.py
billyjacobson: sounds good

billyjacobson (approved):

tseaver:
@chrisrossi Hmm, the newly added samples seem to be flaking, e.g. for PR #306 they [3.6 samples](https://source.cloud.google.com/results/invocations/7d15a8e6-168b-4961-9464-bbb1376400ad/targets/github%2Fpython-bigtable%2Fsamples%2Finstanceadmin/tests) fail (as do the 3.8 ones).

chrisrossi:
> @chrisrossi Hmm, the newly added samples seem to be flaking, e.g. for PR #306 they [3.6 samples](https://source.cloud.google.com/results/invocations/7d15a8e6-168b-4961-9464-bbb1376400ad/targets/github%2Fpython-bigtable%2Fsamples%2Finstanceadmin/tests) fail (as do the 3.8 ones).

Ok, yep. I think a8f33ac is the culprit. I note in the comments above that it introduces a possible race condition, but propose that it's unlikely we'd encounter it IRL. But now I'm realizing that the tests are probably getting run in parallel for different Python versions, which makes the likelihood of encountering it much higher.

samples/instanceadmin/test_instanceadmin.py
chrisrossi: Looks like the preclean is the wrong approach, because of the race condition it introduces it. I guessed we'd be unlikely to hit it, but guessed wrong. I understand your concern here and can have it try again a few times of a deletion fails, but ultimately, a few might get through and have to be removed manually.

tseaver: Usually we would wrap the "must succeed" teardown stuff in a retry, as in [`docs/snippets.py`](https://github.com/googleapis/python-bigtable/blob/e001812cea997ffb56811195b2d59a8d4266fadc/docs/snippets.py#L101-L108).

chrisrossi:
> @chrisrossi Hmm, the newly added samples seem to be flaking, e.g. for PR #306 they [3.6 samples](https://source.cloud.google.com/results/invocations/7d15a8e6-168b-4961-9464-bbb1376400ad/targets/github%2Fpython-bigtable%2Fsamples%2Finstanceadmin/tests) fail (as do the 3.8 ones).

See fix for this in #309

samples/instanceadmin/test_instanceadmin.py
chrisrossi: Oh, that's great. I hadn't seen that before. In this case, since the failure is hypothetical, I'm not sure which exceptions should be trapped here, as I've never seen a failure.

Issue #335, opened 2021-06-24
Unit test coverage below 100%


tseaver:
```
Name                                                                                         Stmts   Miss Branch BrPart  Cover   Missing
----------------------------------------------------------------------------------------------------------------------------------------
google/cloud/bigtable/__init__.py                                                                6      0      0      0   100%
google/cloud/bigtable/app_profile.py                                                            84      0     18      0   100%
google/cloud/bigtable/backup.py                                                                149      0     24      0   100%
google/cloud/bigtable/batcher.py                                                                33      0     10      0   100%
google/cloud/bigtable/client.py                                                                123      7     30      1    95%   209, 226-243
google/cloud/bigtable/cluster.py                                                                77      0     12      0   100%
google/cloud/bigtable/column_family.py                                                          98      0     30      0   100%
google/cloud/bigtable/encryption_info.py                                                        18      3      4      2    77%   52, 55, 64
google/cloud/bigtable/enums.py                                                                  46      0      0      0   100%
google/cloud/bigtable/error.py                                                                  16      4      2      1    72%   44, 53, 61, 64
google/cloud/bigtable/instance.py                                                              120      0     30      0   100%
google/cloud/bigtable/policy.py                                                                 60      0     18      0   100%
google/cloud/bigtable/row.py                                                                   153      0     28      0   100%
google/cloud/bigtable/row_data.py                                                              272      1     78      1    99%   493
google/cloud/bigtable/row_filters.py                                                           236      0     76      0   100%
google/cloud/bigtable/row_set.py                                                                65      0     24      0   100%
google/cloud/bigtable/table.py                                                                 278      2     76      0    99%   1128-1132
google/cloud/bigtable_admin_v2/__init__.py                                                      70      0      0      0   100%
google/cloud/bigtable_admin_v2/services/__init__.py                                              0      0      0      0   100%
google/cloud/bigtable_admin_v2/services/bigtable_instance_admin/__init__.py                      3      0      0      0   100%
google/cloud/bigtable_admin_v2/services/bigtable_instance_admin/async_client.py                285      0     92      0   100%
google/cloud/bigtable_admin_v2/services/bigtable_instance_admin/client.py                      414      0    166      0   100%
google/cloud/bigtable_admin_v2/services/bigtable_instance_admin/pagers.py                       42      0     10      0   100%
google/cloud/bigtable_admin_v2/services/bigtable_instance_admin/transports/__init__.py           9      0      0      0   100%
[...]

PR #337, opened 2021-06-24
tests: restore coverage for 'google.cloud.bigtable.table' to 100%

tseaver:
Toward #335.

### DIFF
--- a/tests/unit/test_table.py
+++ b/tests/unit/test_table.py
@@ -1720,8 +1720,10 @@ class Test__RetryableMutateRowsWorker(unittest.TestCase):
         prior_statuses=None,
         expected_result=None,
         raising_retry=False,
+        retryable_error=False,
         timeout=None,
     ):
+        from google.api_core.exceptions import ServiceUnavailable
         from google.cloud.bigtable.row import DirectRow
         from google.cloud.bigtable.table import _BigtableRetryableError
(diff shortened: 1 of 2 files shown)


tests/unit/test_table.py
```diff
@@ -1711,31 +1713,21 @@ def test_callable_retry(self):
...
-        instance = client.instance(instance_id=self.INSTANCE_ID)
-        table = self._make_table(self.TABLE_ID, instance)
-
-        worker = self._make_worker(client, table.name, [])
-        statuses = worker._do_mutate_retryable_rows()
-
-        self.assertEqual(len(statuses), 0)
-
-    def test_do_mutate_retryable_rows(self):
+    def _do_mutate_retryable_rows_helper(
```
chrisrossi, line 1716: This seems to be the same pattern `@pytest.mark.parametrize` seems to be trying to make possible. Would that be usable/useful here?

tseaver: `parametrize` is about pushing different data into the function, but a loses some of the documentation value (see the comments below in the functions which call the helper).  *Some* of the arguments here are really funky wires to control the flow, versus 'normal" data to pass to the method-under-test.

chrisrossi: Ok!

chrisrossi (approved):

PR #339, opened 2021-06-24
tests: restore assertion for API call

tseaver:
Also fixes coverage (line 505 was unreachable).

Toward #335.


chrisrossi (approved):
LGTM!

PR #343, opened 2021-06-25
tests: restore 100% unit test coverage for 'google.cloud.bigtable.client'

tseaver:
- Tweak some docstrings for PEP 257 conformance.
- Declare gRPC options at module scope
- Clarify logic flow in `Client._create_gapic_client_channel`.

Toward #335.

### DIFF
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -250,6 +250,41 @@ class TestClient(unittest.TestCase):
             emulator_host, lcc.return_value, options=options,
         )
 
+    def test__local_composite_credentials(self):
+        client = self._make_one(
+            project=self.PROJECT, credentials=_make_credentials(), read_only=True
+        )
+
+        wsir_patch = mock.patch("google.auth.credentials.with_scopes_if_required")
+        request_patch = mock.patch("google.auth.transport.requests.Request")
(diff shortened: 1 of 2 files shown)


tests/unit/test_client.py
```diff
@@ -215,6 +212,140 @@ def test__get_scopes_read_only(self):
...
+        wsir_patched.assert_called_once_with(client._credentials, None)
+        request_patched.assert_called_once_with()
+        amp_patched.assert_called_once_with(
+            wsir_patched.return_value, request_patched.return_value,
+        )
+        grpc_mcc.assert_called_once_with(amp_patched.return_value)
+        grpc_lcc.assert_called_once_with()
+        grpc_ccc.assert_called_once_with(grpc_lcc.return_value, grpc_mcc.return_value)
+
+    def _create_gapic_client_channel_helper(
```
chrisrossi, line 288: Same question as I had for #337 re: parametrize.

tseaver: Doesn't feel right here, either:  these params are about wriggling wires, rather than data to pass to the MUT.

chrisrossi: Ok!

chrisrossi (approved):

PR #359, opened 2021-07-09
fix: use public 'table_admin_client' property in backups methods

kolea2:
calling `_table_admin_client` was lazily evaluated and would return a `NoneType` in cases where table operations were not already done.

### DIFF
--- a/tests/unit/test_backup.py
+++ b/tests/unit/test_backup.py
@@ -494,7 +494,7 @@ class TestBackup(unittest.TestCase):
         from google.api_core.exceptions import Unknown
 
         client = _Client()
-        api = client._table_admin_client = self._make_table_admin_client()
+        api = client.table_admin_client = self._make_table_admin_client()
         api.get_backup.side_effect = Unknown("testing")
 
         instance = _Instance(self.INSTANCE_NAME, client=client)
@@ -510,7 +510,7 @@ class TestBackup(unittest.TestCase):
         from google.api_core.exceptions import NotFound
 
         client = _Client()
-        api = client._table_admin_client = self._make_table_admin_client()
+        api = client.table_admin_client = self._make_table_admin_client()
         api.get_backup.side_effect = NotFound("testing")
 
         instance = _Instance(self.INSTANCE_NAME, client=client)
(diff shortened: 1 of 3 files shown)


kolea2:
@crwilcox any ideas on how to reliably test this?

tseaver:
Because a Python property looks just like an attribute, there isn't any good way to distinguish them.  You could add a property to the dummy `_Client` class which returned the value of a *different* attribute, e.g.:

```python


class _Client(object):
    _different_tac = None

    def __init__(self, project=TestBackup.PROJECT_ID):
        self.project = project
        self.project_name = "projects/" + self.project

    @property
    def table_admin_client(self):
        return self._different_tac
```
and then populate `_different_tac` in the tests.

tseaver:
I just checked, and there is a [`mock.PropertyMock`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.PropertyMock) which might do, but it is too gnarly to use IMHO.

tests/unit/test_backup.py
```diff
@@ -494,7 +494,7 @@ def test_exists_grpc_error(self):
         from google.api_core.exceptions import Unknown
```
tseaver, line 494: These methods all need to be updated the same way:
- `test_create_*`,
- `test_get`
- `test_reload`
- `test_delete_*`
-  `test_update_*`
- `test_restore_*`
- `test_get_iam_policy`
- `test_set_iam_policy`
- `test_test_iam_permissions`

tseaver (dismissed):

kolea2:
> Because a Python property looks just like an attribute, there isn't any good way to distinguish them. You could add a property to the dummy `_Client` class which returned the value of a _different_ attribute, e.g.:
> 
> ```python
> class _Client(object):
>     _different_tac = None
> 
>     def __init__(self, project=TestBackup.PROJECT_ID):
>         self.project = project
>         self.project_name = "projects/" + self.project
> 
>     @property
>     def table_admin_client(self):
>         return self._different_tac
> ```
> 
> and then populate `_different_tac` in the tests.

@tseaver discussed with @crwilcox, we decided to go with a slightly different approach here. In the backups system test, reset the Client so that we can assert the table admin client object is initialized and returned. I tested this with and without my patch to confirm this works. LMK what you think!

tseaver:
I don't think I grok the need for the system test change.  If you want to ensure that all the methods which use the table admin client use the property, rather than the attribute, then just setting the same-named attribute on the `_Client` shim in the unit tests should do the trick:  any methods which try to use `_table_admin_client` will fail.

crwilcox:
The system test change ensure we start with all lazy-loading reset.

As for testing, the issue is we mock that bit out as best I can tell. The changes @kolea2  made means the mock no longer even has the internal surface, so it def. isn't being called.

crwilcox (approved):

PR #374, opened 2021-07-14
chore: 'requirements.txt' is not a script

tseaver:


parthea (approved):

tseaver:
@parthea Looks like we need approval from a member of @googleapis/api-bigtable .

kolea2 (approved):

PR #382, opened 2021-07-20
fix: directly append to pb for better read row performance

crwilcox:
Interacting directly with _pb rather than proto-plus gives a significant performance advantage.

### DIFF
--- a/google/cloud/bigtable/row_set.py
+++ b/google/cloud/bigtable/row_set.py
@@ -137,7 +137,7 @@ class RowSet(object):
         :param message: The ``ReadRowsRequest`` protobuf
         """
         for each in self.row_keys:
-            message.rows.row_keys.append(_to_bytes(each))
+            message.rows.row_keys._pb.append(_to_bytes(each))
 
         for each in self.row_ranges:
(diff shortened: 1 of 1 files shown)


google/cloud/bigtable/row_set.py
```diff
@@ -137,7 +137,7 @@ def _update_message_request(self, message):
         :param message: The ``ReadRowsRequest`` protobuf
         """
         for each in self.row_keys:
-            message.rows.row_keys.append(_to_bytes(each))
+            message.rows.row_keys._pb.append(_to_bytes(each))
```
kolea2, line 140: is it at all possible we could run into the same issues we same here with lazy eval? https://github.com/googleapis/python-bigtable/pull/359

crwilcox: I wouldn't expect so. We have applied a similar patch to firestore and pubsub libraries. `row_keys.append` calls proto plus which ultimately will interact with `_pb`. The way it does this though is not fast. :(

kolea2 (approved):

tseaver (approved):

PR #386, opened 2021-07-20
chore: configure renovate to use 'chore(deps)' for its commits

tseaver:
Avoids useless changelog entries for bumping samples to the *previous* version.  See PR #378.


tseaver:
I'm not fighting with Owlbot over this change.

PR #403, opened 2021-08-06
fix: mtls

arithmetic1728:

### DIFF
--- a/google/cloud/bigtable/client.py
+++ b/google/cloud/bigtable/client.py
@@ -258,10 +258,11 @@ class Client(ClientWithProject):
     def _create_gapic_client_channel(self, client_class, grpc_transport):
         if self._emulator_host is not None:
             api_endpoint = self._emulator_host
-        elif self._client_options and self._client_options.api_endpoint:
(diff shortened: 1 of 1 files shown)


google/cloud/bigtable/client.py
```diff
@@ -258,10 +258,11 @@ def _local_composite_credentials(self):
...
         if self._emulator_host is not None:
             api_endpoint = self._emulator_host
-        elif self._client_options and self._client_options.api_endpoint:
-            api_endpoint = self._client_options.api_endpoint
         else:
-            api_endpoint = client_class.DEFAULT_ENDPOINT
+            # let the auto-generated client
+            client = client_class(client_options=self._client_options)
+            api_endpoint = client.transport._host
+            ssl_credentials = client.transport._ssl_channel_credentials
```
tseaver, line 265: This change doesn't smell right:  we are in the middle of creating the gapic client here:  we shouldn't be pre-creating it to get arguments to build the transport to then create another one.  This dance suggests that the constructor for the generated client should be broken up into separate classmethod for finding the endpoint and constructing the transport.

PR #448, opened 2021-10-13
chore: ensure mypy passes

crwilcox:
related to #447 

```
mypy -p google.cloud.bigtable --no-incremental
Success: no issues found in 17 source files
```

For bigtable, specifying a subfolder fails, but package works fine. Ignoring for now
```
❯ ./.nox/unit-3-10/bin/mypy google/cloud/bigtable
google/cloud/bigtable/__init__.py: error: Source file found twice under different module names: "cloud.bigtable.client" and "google.cloud.bigtable.client"
Found 1 error in 1 file (errors prevented further checking)
```

### DIFF
--- a/google/cloud/bigtable/__init__.py
+++ b/google/cloud/bigtable/__init__.py
@@ -15,8 +15,10 @@
 """Google Cloud Bigtable API package."""
 
 
+from typing import Optional
 import pkg_resources  # type: ignore
 
+__version__ : Optional[str]
 try:
     __version__ = pkg_resources.get_distribution("google-cloud-bigtable").version
 except pkg_resources.DistributionNotFound:
--- a/google/cloud/bigtable/__init__.py
+++ b/google/cloud/bigtable/__init__.py
@@ -18,14 +18,13 @@
 from typing import Optional
 import pkg_resources
 
+from google.cloud.bigtable.client import Client
+
 __version__: Optional[str]
 try:
     __version__ = pkg_resources.get_distribution("google-cloud-bigtable").version
 except pkg_resources.DistributionNotFound:
     __version__ = None
-
-
-from google.cloud.bigtable.client import Client
 
 
 __all__ = ["__version__", "Client"]
--- a/google/cloud/bigtable/__init__.py
+++ b/google/cloud/bigtable/__init__.py
@@ -15,7 +15,7 @@
 """Google Cloud Bigtable API package."""
 
 
-import pkg_resources # type: ignore
+import pkg_resources  # type: ignore
 
 try:
     __version__ = pkg_resources.get_distribution("google-cloud-bigtable").version
--- a/google/cloud/bigtable/__init__.py
+++ b/google/cloud/bigtable/__init__.py
@@ -16,9 +16,9 @@
 
 
 from typing import Optional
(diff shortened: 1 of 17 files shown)


tseaver (changes requested):
Shouldn't we be adding a `mypy` nox session?

google/cloud/bigtable/__init__.py
```diff
@@ -15,7 +15,7 @@
 """Google Cloud Bigtable API package."""
 
 
-import pkg_resources
+import pkg_resources  # type: ignore
 
 try:
     __version__ = pkg_resources.get_distribution("google-cloud-bigtable").version
```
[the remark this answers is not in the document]
crwilcox: mostly this yeah. committed a change to this tune.

tseaver (approved):

google/cloud/bigtable/client.py
```diff
@@ -29,11 +29,11 @@
 """
 import os
 import warnings
-import grpc
+import grpc  # type: ignore
 
-from google.api_core.gapic_v1 import client_info
-import google.auth
-from google.auth.credentials import AnonymousCredentials
+from google.api_core.gapic_v1 import client_info  # type: ignore
```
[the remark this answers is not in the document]
crwilcox: I had to make an edit to how we used exceptions as it caused an error that was rough to get rid of. We should be able to get rid of some of these though, yes.

google/cloud/bigtable/table.py
```diff
@@ -1043,10 +1043,6 @@ class _RetryableMutateRowsWorker(object):
     are retryable, any subsequent call on this callable will be a no-op.
     """
 
-    RETRY_CODES = tuple(
```
crwilcox, line 1046: FYI, this will give `google/cloud/bigtable/table.py:1047: error: "None" has no attribute "value"` and even if you add an `if` that would validate non-none, it will persist. If we needed this const here (only used in one spot) we would have to likely do in a for loop, not comprehension. Possibly a shortcoming of mypy

google/cloud/bigtable/table.py
```diff
@@ -1083,7 +1079,12 @@ def __call__(self, retry=DEFAULT_RETRY):
 
     @staticmethod
     def _is_retryable(status):
-        return status is None or status.code in _RetryableMutateRowsWorker.RETRY_CODES
+        RETRY_CODES = tuple(
+            retryable.grpc_status_code.value[0]
+            for retryable in RETRYABLE_MUTATION_ERRORS
+        )
```
tseaver, line 1085: Seems like a performance hit here to recompute every time.

google/cloud/bigtable/table.py
tseaver: Could we appease it by adding `if retryable.grpc_status_code is not None`?

google/cloud/bigtable/table.py
crwilcox: Yeah. This is a bit irritating. So I tried adding an if clause to the list comp, but the error was still there. If I structured it as a regular for loop to a list, then cast to tuple, it likely will pass, but that is a bit.. gross?

tseaver:
@googlebot I consent

PR #509, opened 2022-02-15
feat: Add support for autoscaling

Mariatta:
- Add the parameters min_serve_nodes, max_serve_nodes, and cpu_utilization_percent
- Create disable_autoscaling function
- Update documentation and tests

### DIFF
--- a/google/cloud/bigtable/cluster.py
+++ b/google/cloud/bigtable/cluster.py
@@ -263,13 +263,13 @@ class Cluster(object):
             and not self.cpu_utilization_percent
         ):
             raise ValueError(
-                "Must specify either serve_nodes or all of the autoscaling configurations."
+                "Must specify either serve_nodes or all of the autoscaling configurations (min_serve_nodes, max_serve_nodes, and cpu_utilization_percent)."
             )
         if self.serve_nodes and (
             self.max_serve_nodes or self.min_serve_nodes or self.cpu_utilization_percent
         ):
             raise ValueError(
-                "Cannot specify both serve_nodes and autoscaling configurations."
+                "Cannot specify both serve_nodes and autoscaling configurations (min_serve_nodes, max_serve_nodes, and cpu_utilization_percent)."
             )
         if (
             (
@@ -286,7 +286,7 @@ class Cluster(object):
             )
         ):
             raise ValueError(
-                "All of autoscaling configurations must be specified at the same time."
+                "All of autoscaling configurations must be specified at the same time (min_serve_nodes, max_serve_nodes, and cpu_utilization_percent)."
             )
 
     def __eq__(self, other):
--- a/google/cloud/bigtable/cluster.py
+++ b/google/cloud/bigtable/cluster.py
@@ -18,6 +18,7 @@
 import re
 from google.cloud.bigtable_admin_v2.types import instance
 from google.api_core.exceptions import NotFound
+from google.protobuf import field_mask_pb2
 
 
 _CLUSTER_NAME_RE = re.compile(
@@ -36,6 +37,7 @@ class Cluster(object):
     * :meth:`create` itself
     * :meth:`update` itself
     * :meth:`delete` itself
+    * :meth:`disable_autoscaling` itself
 
     :type cluster_id: str
     :param cluster_id: The ID of the cluster.
@@ -52,7 +54,9 @@ class Cluster(object):
                         https://cloud.google.com/bigtable/docs/locations
 
     :type serve_nodes: int
-    :param serve_nodes: (Optional) The number of nodes in the cluster.
+    :param serve_nodes: (Optional) The number of nodes in the cluster. If any, of the
+                        autoscaling configuration are specified, then the autoscaling
+                        configuration will take precedent.
 
     :type default_storage_type: int
     :param default_storage_type: (Optional) The type of storage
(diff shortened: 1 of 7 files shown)


google/cloud/bigtable/cluster.py
```diff
@@ -85,6 +89,28 @@ class Cluster(object):
                    :data:`google.cloud.bigtable.enums.Cluster.State.CREATING`.
                    :data:`google.cloud.bigtable.enums.Cluster.State.RESIZING`.
                    :data:`google.cloud.bigtable.enums.Cluster.State.DISABLED`.
+
+    :type min_serve_nodes: int
+    :param min_serve_nodes: (Optional) The minimum number of nodes in the cluster. Must be
```
kolea2, line 94: I would reword this slightly to point out that this is a setting for autoscaling. Something like:
```
The miniumum number of nodes to be set in the cluster for autoscaling. Must be 1 or greater ...
```

google/cloud/bigtable/cluster.py
```diff
@@ -52,7 +54,9 @@ class Cluster(object):
                         https://cloud.google.com/bigtable/docs/locations
 
     :type serve_nodes: int
-    :param serve_nodes: (Optional) The number of nodes in the cluster.
+    :param serve_nodes: (Optional) The number of nodes in the cluster. If any, of the
```
kolea2, line 57: Slight rewording suggestion:
```
The number of nodes in the cluster for manual scaling.
```

google/cloud/bigtable/cluster.py
```diff
@@ -104,10 +133,13 @@ def __init__(
         self.default_storage_type = default_storage_type
         self._kms_key_name = kms_key_name
         self._state = _state
+        self.min_serve_nodes = min_serve_nodes
```
kolea2, line 136: do you think it would be a good idea to do some validation here? Here are some scenarios that come to mind:
- The user sets serve nodes AND autoscaling parameters
- The user sets no settings relating to auto or manual scaling
- The user only sets one or two (not all three) autoscaling config parameters

google/cloud/bigtable/cluster.py
```diff
@@ -323,20 +366,61 @@ def update(self):
...
-                "location": None,
-            }
+
+        update_mask_pb = field_mask_pb2.FieldMask()
+
+        if self.serve_nodes:
+            update_mask_pb.paths.append("serve_nodes")
+
+        if self.min_serve_nodes or self.max_serve_nodes or self.cpu_utilization_percent:
+            update_mask_pb.paths.append("cluster_config")
```
kolea2, line 384: I think we should add the more granular path - `cluster_config.cluster_autoscaling_config` and update `_to_pb()` to reflect that

tests/system/test_instance_admin.py
```diff
@@ -614,3 +652,37 @@ def test_cluster_update(
     admin_cluster.serve_nodes = serve_nodes
     operation = admin_cluster.update()
     operation.result(timeout=60)  # Ensure the operation completes.
+
+
+def test_cluster_update_w_autoscaling(
```
kolea2, line 657: looks good - I would also add a test going from auto -> manual scaling (disabling autoscaling)

kolea2 (changes requested):

tests/system/test_instance_admin.py
Mariatta: I added `test_cluster_disable_autoscaling` below.

google/cloud/bigtable/cluster.py
Mariatta: Updated.

google/cloud/bigtable/cluster.py
Mariatta: Updated.

google/cloud/bigtable/cluster.py
Mariatta: Great suggestion. I added the `_validate_scaling_config` function, which gets called in `create` and `update`. I also added unit tests to test the above scenarios.

What do you think about this?

google/cloud/bigtable/cluster.py
```diff
@@ -211,6 +253,42 @@ def kms_key_name(self):
...
+        """Validate auto/manual scaling configuration before creating or updating."""
+
+        if (
+            not self.serve_nodes
+            and not self.min_serve_nodes
+            and not self.max_serve_nodes
+            and not self.cpu_utilization_percent
+        ):
+            raise ValueError(
+                "Must specify either serve_nodes or all of the autoscaling configurations."
```
kolea2, line 266: maybe list the configurations here for convenience?

kolea2 (commented):
lgtm after two small docs comments!

tests/system/test_instance_admin.py
```diff
@@ -614,3 +652,69 @@ def test_cluster_update(
...
+
+    operation = admin_cluster_with_autoscaling.update()
+    operation.result(timeout=60)  # Ensure the operation completes.
+
+    # Create a new cluster instance and reload it.
+    alt_cluster = admin_instance_populated.cluster(admin_cluster_id)
+    alt_cluster.reload()
+    assert alt_cluster.min_serve_nodes == new_min_serve_nodes
+    assert alt_cluster.max_serve_nodes == new_max_serve_nodes
+    assert alt_cluster.cpu_utilization_percent == new_cpu_utilization_percent
```
kolea2, line 681: let's add one more test case in which only one field is updated.

kolea2 (changes requested):

tests/system/test_instance_admin.py
Mariatta: I've added such test below (line 691)

kolea2 (approved):

kolea2 (approved):

PR #512, opened 2022-02-17
fix: Pass app_profile_id when building updated request

Mariatta:
- Pass the app_profile_id from the message
- Update unittest

Fixes internal bug #214449800

### DIFF
--- a/google/cloud/bigtable/row_data.py
+++ b/google/cloud/bigtable/row_data.py
@@ -453,6 +453,7 @@ class PartialRowsData(object):
         req_manager = _ReadRowsRequestManager(
             self.request, self.last_scanned_row_key, self._counter
         )
+        print("retry request")
         return req_manager.build_updated_request()
 
     def _on_error(self, exc):
@@ -627,6 +628,7 @@ class _ReadRowsRequestManager(object):
         r_kwargs = {
             "table_name": self.message.table_name,
             "filter": self.message.filter,
+            "app_profile_id": self.message.app_profile_id,
         }
 
         if self.message.rows_limit != 0:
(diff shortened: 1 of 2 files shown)


google/cloud/bigtable/row_data.py
```diff
@@ -627,6 +628,7 @@ def build_updated_request(self):
         r_kwargs = {
             "table_name": self.message.table_name,
             "filter": self.message.filter,
+            "app_profile_id": self.message.app_profile_id,
```
igorbernstein2, line 630: I think this works, but it puts us in the same position as before: if ReadRowsRequest ever adds a new field, we will end up dropping that field on retry. I think one of 2 things should happen:
1. clone the entry request and overwrite the RowSet & limit. This way if new fields are added, we are sure to copy them
2. Add a unit test that ensure no new fields are added with a comment to instructing someone to update this code

I think 1 is definitely preferable

igorbernstein2 (changes requested):

kolea2 (approved):

tests/unit/test_row_data.py
```diff
@@ -902,7 +905,10 @@ def test_RRRM_build_updated_request(rrrm_data):
     result = request_manager.build_updated_request()
 
     expected_result = _ReadRowsRequestPB(
-        table_name=TABLE_NAME, filter=row_filter.to_pb(), rows_limit=6
+        table_name=TABLE_NAME,
```
kolea2, line 908: is it easy to also check this in the system tests?

Mariatta: I just ran the `system` test suite, but the ``build_updated_request`` doesn't seem to be covered there currently.

igorbernstein2 (approved):
lgtm

PR #533, opened 2022-03-10
fix: Ensure message fields are copied when building retry request

Mariatta:
Use the copy_from() function when building the ReadRowsRequest arguments.

Fixes internal bug #214449800

Fixes #214449800 🦕

### DIFF
--- a/google/cloud/bigtable/row_data.py
+++ b/google/cloud/bigtable/row_data.py
@@ -16,6 +16,7 @@
 
 
 import copy
+import proto
 
 import grpc  # type: ignore
 
@@ -626,7 +627,7 @@ class _ReadRowsRequestManager(object):
         """Updates the given message request as per last scanned key"""
 
         updated_message = data_messages_v2_pb2.ReadRowsRequest()
-        data_messages_v2_pb2.ReadRowsRequest.copy_from(updated_message, self.message)
+        proto.Message.copy_from(updated_message, self.message)
 
         if self.message.rows_limit != 0:
             updated_message.rows_limit = max(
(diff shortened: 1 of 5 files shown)


google/cloud/bigtable/row_data.py
```diff
@@ -625,16 +626,11 @@ def __init__(self, message, last_scanned_key, rows_read_so_far):
...
         """Updates the given message request as per last scanned key"""
 
-        # TODO: Generalize this to ensure fields don't get rewritten when retrying the request
-
-        r_kwargs = {
-            "table_name": self.message.table_name,
-            "filter": self.message.filter,
-            "app_profile_id": self.message.app_profile_id,
-        }
+        updated_message = data_messages_v2_pb2.ReadRowsRequest()
```
igorbernstein2, line 629: I would call this a `resume_request`

google/cloud/bigtable/row_data.py
```diff
@@ -625,16 +626,11 @@ def __init__(self, message, last_scanned_key, rows_read_so_far):
...
+        proto.Message.copy_from(updated_message, self.message)
 
         if self.message.rows_limit != 0:
-            r_kwargs["rows_limit"] = max(
+            updated_message.rows_limit = max(
                 1, self.message.rows_limit - self.rows_read_so_far
```
igorbernstein2, line 634: why max? If the `rows_limit - self.rows_read_so_far` is negative then something is terribly wrong and we should throw an error. If its 0 then then stream should be completed before we even get here. Please add a TODO: to address this.

igorbernstein2 (changes requested):
It looks like unit tests are broken. Please fix

google/cloud/bigtable/row_data.py
Mariatta: I've added the TODO, and the tests have been fixed.

google/cloud/bigtable/row_data.py
Mariatta: Yup, updated.

igorbernstein2 (approved):
lgtm

PR #549, opened 2022-03-31
feat: App Profile multi cluster routing support with specified cluster ids

Mariatta:
Add the `multi_cluster_ids` parameter to AppProfile.
It is used when the routing policy type is set to "any".

Add tests, docstrings.
Fix some typos.

Closes internal issue 213627978

Fixes #213627978 🦕

### DIFF
--- a/tests/system/test_instance_admin.py
+++ b/tests/system/test_instance_admin.py
@@ -438,7 +438,7 @@ def test_instance_create_app_profile_create_with_multi_cluster_ids(
     )
     operation = instance.create(clusters=[cluster_1, cluster_2])
     instances_to_delete.append(instance)
-    operation.result(timeout=120)  # Ensure the operation completes.
+    operation.result(timeout=240)  # Ensure the operation completes.
 
     # Create a new instance and make sure it is the same.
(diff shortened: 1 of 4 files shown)


parthea (approved):

tests/system/test_instance_admin.py
```diff
@@ -87,6 +91,7 @@ def _modify_app_profile_helper(
     assert alt_profile.description == description
     assert alt_profile.routing_policy_type == routing_policy_type
     assert alt_profile.cluster_id == cluster_id
+    assert alt_profile.multi_cluster_ids == multi_cluster_ids
```
kolea2, line 94: can we add a few more system tests that create an app profile with multi cluster routing and ids specified, and also an update test adding or removing a cluster id?

kolea2 (changes requested):

tests/system/test_instance_admin.py
Mariatta: I've added the test below: `test_instance_create_app_profile_create_with_multi_cluster_ids`

kolea2 (approved):

Mariatta:
Thanks!

PR #602, opened 2022-06-07
fix(deps): require protobuf<4.0.0 on v1 branch

parthea:


Mariatta:
The `docs` build is failing because of a recent release of Sphinx v 5.
I made a separate PR to address just the docs build: https://github.com/googleapis/python-bigtable/pull/603/files

Mariatta:
We can also alternatively pin it to Sphinx <5 but I think keeping it up to date and recent would be better.

Mariatta (approved):

PR #623, opened 2022-08-05
test: improve the row merging tests

igorbernstein2:
- extract read-rows-acceptance-test.json based tests into own file
- update the json to match the latest available in https://github.com/googleapis/conformance-tests/tree/main/bigtable/v2
- use parameterized pytest test to run all of the scenarios (instead of creating a function for each json blob)
- use json protobufs to parse the file

I left a TODO to allow easy updates of the file, unfortunately its not straight forward as the canonical protos get renamed for python gapic
Next PR will extract row merging functionality from row_data to make it easier to maintain


Mariatta (approved):
Thank you. I think the refactoring using pytest parametrization is great.

PR #624, opened 2022-08-05
Fix: Retry the RST Stream error in mutate rows

Mariatta:
In mutate_rows, Internal Server with RST Stream errors is considered transient, and should be retried.

### DIFF
--- a/google/cloud/bigtable/table.py
+++ b/google/cloud/bigtable/table.py
@@ -67,7 +67,6 @@ RETRYABLE_MUTATION_ERRORS = (
 RETRYABLE_INTERNAL_ERROR_MESSAGES = (
     "rst_stream",
     "rst stream",
-    "received rst stream",
     "received unexpected eos on data frame from server",
 )
 """Internal error messages that can be retried during row mutation."""
@@ -1151,10 +1150,12 @@ class _RetryableMutateRowsWorker(object):
             # For InternalServerError, it is only retriable if the message is related to RST Stream messages
             if (
                 isinstance(exc, InternalServerError)
-                and exc.message.lower() in RETRYABLE_INTERNAL_ERROR_MESSAGES
+                and any(
+                    retryable_message in exc.message.lower()
+                    for retryable_message in RETRYABLE_INTERNAL_ERROR_MESSAGES
+                )
+                or not isinstance(exc, InternalServerError)
             ):
-                raise _BigtableRetryableError
-            elif not isinstance(exc, InternalServerError):
                 raise _BigtableRetryableError
             else:
                 # re-raise the original exception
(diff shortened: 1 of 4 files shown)


google/cloud/bigtable/table.py
```diff
@@ -1130,11 +1143,21 @@ def _do_mutate_retryable_rows(self):
...
-        except RETRYABLE_MUTATION_ERRORS:
+        except RETRYABLE_MUTATION_ERRORS as exc:
             # If an exception, considered retryable by `RETRYABLE_MUTATION_ERRORS`, is
             # returned from the initial call, consider
             # it to be retryable. Wrap as a Bigtable Retryable Error.
-            raise _BigtableRetryableError
+            # For InternalServerError, it is only retriable if the message is related to RST Stream messages
+            if (
+                isinstance(exc, InternalServerError)
+                and exc.message in RETRYABLE_MUTATION_ERRORS
```
mutianf, line 1153: can we do exc.message.toLowerCase and update RETRYABLE_MUTATION_ERRORS to have {rst_stream, rst stream}? Because the error message might be slightly different, but "rst_stream" and "rst stream" should cover all cases.

Mariatta: I added `.lower()` and `rst stream` to `RETRYABLE_INTERNAL_ERROR_MESSAGES`

tests/unit/test_table.py
```diff
@@ -1785,6 +1792,52 @@ def test_rmrw_do_mutate_retryable_rows_w_retryable_error():
...
+    row_cells = [
+        (b"row_key_1", ("cf", b"col", b"value1")),
+        (b"row_key_2", ("cf", b"col", b"value2")),
+    ]
+    responses = ()
+
+    for retryable_internal_error_message in RETRYABLE_INTERNAL_ERROR_MESSAGES:
+        for message in [
+            retryable_internal_error_message,
+            retryable_internal_error_message.upper(),
```
Mariatta, line 1811: I added `upper` to test that it's case insensitive.

google/cloud/bigtable/table.py
```diff
@@ -55,9 +56,22 @@
...
+    DeadlineExceeded,
+    ServiceUnavailable,
+    InternalServerError,
+)
 """Errors which can be retried during row mutation."""
 
+RETRYABLE_INTERNAL_ERROR_MESSAGES = (
+    "rst_stream",
+    "rst stream",
+    "received rst stream",
```
mutianf, line 70: I don't think you need this because it's included in the line above

google/cloud/bigtable/table.py
```diff
@@ -1130,11 +1143,23 @@ def _do_mutate_retryable_rows(self):
                 retry=None,
                 **kwargs
             )
-        except RETRYABLE_MUTATION_ERRORS:
+        except RETRYABLE_MUTATION_ERRORS as exc:
```
mutianf, line 1146: It should also be retried in read rows

Mariatta: I've updated `row_data` module to also handle rst stream retry read rows.

mutianf (approved):

PR #680, opened 2022-09-27
feat: Copy Backup

Mariatta:


google/cloud/bigtable/backup.py
```diff
@@ -339,6 +356,68 @@ def create(self, cluster_id=None):
...
+        cluster_id=None,
+        expire_time=None,
+    ):
+        """Make a copy of this backup.
+
+        :type new_backup_id: str
+        :param new_backup_id: The name of the copied backup
+
+        :type project_id: str
+        :param project_id: The destination project id for the backup
```
kolea2, line 373: is this also optional? same question for other parameters

google/cloud/bigtable/backup.py
```diff
@@ -339,6 +356,68 @@ def create(self, cluster_id=None):
...
+
+        parent = BigtableTableAdminClient.cluster_path(
+            project=project_id,
+            instance=instance_id,
+            cluster=cluster_id,
+        )
+
+        return api.copy_backup(
+            request={
+                "parent": parent,
```
kolea2, line 414: how does this work if the source backup project and parent backup project are different?

tests/system/test_table_admin.py
```diff
@@ -323,6 +324,27 @@ def test_table_backup(
         temp_table_backup.expire_time
     )
 
+    # Testing `backup.copy()` method
+    backup_op = temp_backup.copy("copied-backup")
```
kolea2, line 328: I would also add some test cases with some of the other parameters set.

kolea2 (changes requested):

google/cloud/bigtable/backup.py
Mariatta: `parent` is constructed using `project_id`, `instance_id`, `cluster_id`. It's allowed to be different I believe as long as the user have access to both. I'm not seeing any restriction about it from the design doc. From the client we will pass the path, and relay back error from the server.

tests/system/test_table_admin.py
Mariatta: I've added tests that pass additional parameters.

tests/system/test_table_admin.py
```diff
@@ -354,3 +350,103 @@ def test_table_backup(
...
+    temp_backup = source_table.backup(
+        temp_backup_id,
+        cluster_id=data_cluster_id,
+        expire_time=expire,
+    )
+    temp_backup.create()
+    backups_to_delete.append(temp_backup)
+
+    # Test copying to the same instance
+    backup_op = temp_backup.copy("copied-backup")
```
kolea2, line 393: let's add a unique identifier to `copied-backup`

tests/system/test_table_admin.py
```diff
@@ -354,3 +350,103 @@ def test_table_backup(
...
+    alt_cluster = alt_instance.cluster(
+        cluster_id=alt_cluster_id,
+        location_id=location_id,
+        serve_nodes=1,
+    )
+    alt_instance.create(clusters=[alt_cluster])
+    instances_to_delete.append(alt_instance)
+
+    # create a copy to the new instance, cluster, and new expire_time
+    copied_backup_id = "copy-new-instance"
```
kolea2, line 424: same here

tests/system/test_table_admin.py
Mariatta: I've added `-{unique_suffix}` to the copy backup ids

PR #686, opened 2022-10-20
fix: workaround to use default deadlines

igorbernstein2:
This is a workaround for https://github.com/googleapis/gapic-generator-python/issues/1477.

The underlying gapic client in the python-bigtable has a bug that prevents it from using the default client deadlines configured in base.py. The root cause is that the gapic micro generator sets the default deadline = None in rpc method signatures. This signals to the underlying infrastructure to strip away any default deadlines.

This PR monkeypatches the method signatures to use default deadlines by default by passing in the default sentinel value


vam-google (changes requested):
Please just wait for this being fixed in gapic-gen.

PR #695, opened 2022-11-17
fix: first pass on making retry configuration more consistent

igorbernstein2:
Currently ReadRows uses the Retry.deadline configuration inconsistently:
- its used as the attempt timeout for the first retry attempt
- its used as a limit for retry scheduling for reading a single row

Conceptually there are 3 timeouts that are relevant to ReadRows:
- attempt timeout: how long a single RPC is allowed to run, this should map directly to a gRPC deadline
- overall timeout: Limit how long we should wait across all of the retry attempts, possibly truncating the last attempt timeout.
- read timeout: How long we are willing to wait for the next row in a stream

Ideally Retry.deadline would represent an operation deadline (since thats the primary concern of the end user). However there is no backwards compatible way to do this. Changing the behavior would cause existing application to start enforcing a very short deadline.

This PR tries to improve the situation in a backwards compatible way:
- keep Retry.deadline as a read timeout
- introduce a new parameter for overall timeout

This results in less than ideal api, but avoids breaking existing applications.

### DIFF
--- a/google/cloud/bigtable/row_data.py
+++ b/google/cloud/bigtable/row_data.py
@@ -509,13 +509,16 @@ class PartialRowsData(object):
         # Calculate the maximum amount of time that retries should be scheduled.
         # This will not actually set any deadlines, it will only limit the
         # duration of time that we are willing to schedule retries for.
-        effective_timeout = self.remaining_overall_timeout
+        remaining_overall_timeout = self.remaining_overall_timeout
 
-        if self.retry.deadline is not None:
-            if effective_timeout is None or self.retry.deadline < effective_timeout:
-                effective_timeout = self.retry.deadline
-
-        self.retry = self.retry.with_deadline(effective_timeout)
+        if remaining_overall_timeout is not None:
+            # we want make sure that the retry logic doesnt retry after the
+            # operation deadline is past
+            if (
+                self.retry.deadline is None
+                or self.retry.deadline > remaining_overall_timeout
+            ):
+                self.retry = self.retry.with_deadline(remaining_overall_timeout)
 
         return self.retry(self._read_next, on_error=self._on_error)()
 
--- a/google/cloud/bigtable/row_data.py
+++ b/google/cloud/bigtable/row_data.py
@@ -16,6 +16,8 @@
 
 
 import copy
+import time
+
 import six
 
 import grpc
@@ -340,6 +342,10 @@ DEFAULT_RETRY_READ_ROWS = retry.Retry(
     initial=1.0,
     maximum=15.0,
     multiplier=2.0,
+    # NOTE: this is a soft read timeout: this limits for how long we are willing
+    # to schedule retry attempts to read the next row. This does not set the
+    # RPC timeout. Please use the separate overal_timeout parameter of read_rows
+    # to limit the attempt duration
     deadline=60.0,  # 60 seconds
 )
 """The default retry strategy to be used on retry-able errors.
@@ -387,7 +393,9 @@ class PartialRowsData(object):
         STATE_CELL_IN_PROGRESS: CELL_IN_PROGRESS,
     }
 
-    def __init__(self, read_method, request, retry=DEFAULT_RETRY_READ_ROWS):
+    def __init__(
+        self, read_method, request, retry=DEFAULT_RETRY_READ_ROWS, overall_timeout=None
+    ):
         # Counter for rows returned to the user
         self._counter = 0
         # In-progress row, unset until first response, after commit/reset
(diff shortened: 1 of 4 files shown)


google/cloud/bigtable/table.py
```diff
@@ -506,7 +506,9 @@ def read_row(self, row_key, filter_=None):
         """
         row_set = RowSet()
         row_set.add_row_key(row_key)
-        result_iter = iter(self.read_rows(filter_=filter_, row_set=row_set))
+        result_iter = iter(
+            self.read_rows(filter_=filter_, row_set=row_set, overall_timeout=60)
```
mutianf, line 510: is it ok to set a default overall_timeout?

mutianf: Is this the dafult from DEFAULT_RETRY_READ_ROWS.deadline? If so, can we use that instead?

google/cloud/bigtable/table.py
```diff
@@ -615,6 +640,10 @@ def yield_rows(self, **kwargs):
         :param row_set: (Optional) The row set containing multiple row keys and
                         row_ranges.
```
mutianf, line 642: Do we also need to add attempt_timeout here?

igorbernstein2: yep, done!

google/cloud/bigtable/table.py
igorbernstein2: No, this is the default unary deadline from the service json file

google/cloud/bigtable/row_data.py
```diff
@@ -340,6 +342,10 @@ def _retry_read_rows_exception(exc):
     initial=1.0,
     maximum=15.0,
     multiplier=2.0,
+    # NOTE: this is a soft read timeout: this limits for how long we are willing
+    # to schedule retry attempts to read the next row. This does not set the
+    # RPC timeout. Please use the separate overal_timeout parameter of read_rows
+    # to limit the attempt duration
```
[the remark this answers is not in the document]
igorbernstein2: done

google/cloud/bigtable/table.py
Mariatta: Can the `60` be retrieved via configuration or in env var instead of default like this?

tests/unit/test_row_data.py
```diff
@@ -384,20 +384,25 @@ def test_constructor(self):
...
+        partial_rows_data.read_method.assert_called_once_with(request, timeout=mock.ANY)
+
+        # the deadline being passed to the first RPC should be close to 11
+        # But to avoid flakiness on slow test runners, its padded down by 3 secs
+        self.assertLess(8, partial_rows_data.read_method.call_args.kwargs["timeout"])
+
         self.assertIs(partial_rows_data.request, request)
         self.assertEqual(partial_rows_data.rows, {})
-        self.assertEqual(partial_rows_data.retry, retry)
+        # The remaining deadline should be
```
mutianf, line 402: Did you mean
```suggestion
        # The remaining deadline should be less than 11
```

tests/unit/test_row_data.py
```diff
@@ -384,20 +384,25 @@ def test_constructor(self):
...
+        # But to avoid flakiness on slow test runners, its padded down by 3 secs
```
mutianf, line 403: why are there still 2 asserts?

mutianf: Ah I think these few lines can be removed since you already checked on line 398?

tests/unit/test_table.py
```diff
@@ -857,6 +851,74 @@ def test_read_retry_rows(self):
...
+            1.0,
+            client._table_data_client.transport.read_rows.call_args.kwargs["timeout"],
+        )
+
+        # overall timeout should be passed thru
+        client._table_data_client.transport.read_rows = mock.Mock(
+            side_effect=[_MockReadRowsIterator()]
+        )
+        list(table.read_rows(overall_timeout=10.0))
+        self.assertLess(
```
mutianf, line 890: nit: maybe also add a comment about padding slow runners?

tests/unit/test_table.py
```diff
@@ -857,6 +851,74 @@ def test_read_retry_rows(self):
...
+                time.sleep(0.75)
+                raise DeadlineExceeded("delayed error")
+
+            __next__ = next
+
+        client._table_data_client.transport.read_rows = mock.Mock(
+            side_effect=[DelayedFailureIterator(), _MockReadRowsIterator()]
+        )
+        list(table.read_rows(attempt_timeout=1.0, overall_timeout=1.0))
+        self.assertGreater(
```
mutianf, line 917: Hmm why is this greater?

mutianf: are you trying to test where the attempt_timetout > overall_timeout after a few runs?

mutianf (changes requested):

tests/unit/test_table.py
igorbernstein2: No, the assert api is terrible
self.assertGreater(a,b) means

if not a > b:
  fail

so this is asserting that the rpc timeout is less than the attempt timeout

tests/unit/test_table.py
igorbernstein2: done

tests/unit/test_row_data.py
igorbernstein2: No they check different things

tests/unit/test_row_data.py
igorbernstein2: reworded it.

google/cloud/bigtable/table.py
igorbernstein2: i bubbled it up as a kwarg to the method

tests/unit/test_row_data.py
mutianf: Hmm did you commit this change?

mutianf (approved):

PR #696, opened 2022-11-17
fix: make sure that the proper exception type is bubbled up for ReadRows

igorbernstein2:
ReadRows (in client 1.x) bypasses gapic layer and use the raw grpc stubs. Which means that the errors are not being wrapped. Previously the this conversion would be handled in the Retry predicate, however this creates usability issues for end users that are trying to change the exceptions they want to retry on. This PR moves the wrapping a bit lower


mutianf (approved):
Is it possible to add a test on override the exception? Can we document how a customer can override the retry predicate?

PR #554, opened 2022-04-04
fix: Prevent sending full table scan when retrying

Mariatta:
Update the retry logic. Don't send empty row_key and empty row_ranges
if the original message didn't ask for those.

Closes internal issue 214449800

Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigtable/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)

Fixes #<issue_number_goes_here> 🦕

PR #697, opened 2022-11-18
fix: Prevent sending full table scan when retrying (backport #554)

igorbernstein2:
Update the retry logic. Don't send empty row_key and empty row_ranges if the original message didn't ask for those.

NOTE: the test structure changed considerably in the main branch, so backporting those will take some time

### DIFF
--- a/google/cloud/bigtable/row_data.py
+++ b/google/cloud/bigtable/row_data.py
@@ -329,6 +329,10 @@ class InvalidChunk(RuntimeError):
     """Exception raised to to invalid chunk data from back-end."""
 
 
+class InvalidRetryRequest(RuntimeError):
+    """Exception raised when retry request is invalid."""
+
+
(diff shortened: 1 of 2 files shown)


google/cloud/bigtable/row_data.py
```diff
@@ -487,6 +491,9 @@ def __iter__(self):
                 if self.state != self.NEW_ROW:
                     raise ValueError("The row remains partial / is not committed.")
                 break
+            except InvalidRetryRequest:
+                self._cancelled = True
```
mutianf, line 495: Will this show up as cancelled status instead of success?

mutianf (approved):
LGTM

google/cloud/bigtable/row_data.py
igorbernstein2: no it should end in a success...it will just break out of the loop

PR #699, opened 2022-11-18
fix: make internal rst_stream errors retriable

igorbernstein2:

### DIFF
--- a/google/cloud/bigtable/row_data.py
+++ b/google/cloud/bigtable/row_data.py
@@ -18,6 +18,7 @@
 import copy
 import time
 
+import google.api_core.exceptions
 import six
 
 import grpc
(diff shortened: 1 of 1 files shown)


google/cloud/bigtable/row_data.py
```diff
@@ -507,7 +508,16 @@ def _read_next(self):
...
             return six.next(self.response_iterator)
         except grpc.RpcError as grpc_error:
-            raise exceptions.from_grpc_error(grpc_error)
+            # TODO: this needs to be moved to a more general location (ie interceptor)
+            e = exceptions.from_grpc_error(grpc_error)
+            # Sometimes GOAWAYs are surfaced as INTERNAL errors, which makes
+            # them unretriable. This patches that behavior
+            if (
+                e.grpc_status_code == grpc.StatusCode.INTERNAL
+                and "RST_STREAM" in e.message
```
mutianf, line 517: I think the error messages comes in different formats like "Rst stream" "RST_STREAM" etc. maybe do: e.message.tolowercase and check for both "rst stream" and "rst_stream".

igorbernstein2: done

mutianf (approved):

PR #702, opened 2022-11-22
Asynchronous batching

Mariatta:
Call the async mutate_rows from grpc aio

Create flush and mutate_coroutines


google/cloud/bigtable_v2/services/bigtable/async_client.py
```diff
@@ -526,7 +526,7 @@ async def mutate_row(
         # Done; return the response.
         return response
 
-    def mutate_rows(
+    async def mutate_rows(
```
Mariatta, line 529: This needs to be fixed in gapic generator. 
Related PR: https://github.com/googleapis/gapic-generator-python/pull/1502

PR #722, opened 2023-01-12
Feat: Threaded MutationsBatcher

Mariatta:
- Batch mutations in a thread to allow concurrent batching
- Flush the batch every second
- Flow control
- Batcher can now be used as a contextmanager

### DIFF
--- a/google/cloud/bigtable/batcher.py
+++ b/google/cloud/bigtable/batcher.py
@@ -18,7 +18,6 @@ import queue
 import concurrent.futures
 import atexit
 
-from google.cloud.bigtable.error import Status
 
 FLUSH_COUNT = 100
 MAX_MUTATIONS = 100000
@@ -291,9 +290,16 @@ class MutationsBatcher(object):
         """
         response = []
         if len(rows_to_flush) > 0:
-            # returns a list of error codes
+            # returns a list of status codes
             response = self.table.mutate_rows(rows_to_flush)
-            if any(isinstance(result, Status) for result in response):
+            responses = []
+            has_error = False
+            for result in response:
+                if result.code != 0:
+                    has_error = True
+                responses.append(result)
+
+            if has_error:
                 raise MutationsBatchError(status_codes=response)
         return response
 
--- a/google/cloud/bigtable/batcher.py
+++ b/google/cloud/bigtable/batcher.py
@@ -249,7 +249,7 @@ class MutationsBatcher(object):
         self._rows.put(row)
 
         if self._rows.full():
-            self.flush_async()
+            self._flush_async()
 
     def mutate_rows(self, rows):
         """Add multiple rows to the batch. If the current batch meets one of the size
@@ -290,7 +290,7 @@ class MutationsBatcher(object):
         response = self._flush_rows(rows_to_flush)
         return response
 
-    def flush_async(self):
+    def _flush_async(self):
         """Sends the current batch to Cloud Bigtable asynchronously.
 
         :raises:
--- a/google/cloud/bigtable/batcher.py
+++ b/google/cloud/bigtable/batcher.py
@@ -145,7 +145,8 @@ class MutationsBatcher(object):
         self._executor = concurrent.futures.ThreadPoolExecutor()
         self._is_open = True
         atexit.register(self.close)
-        threading.Timer(flush_interval, self.flush).start()
+        self._timer = threading.Timer(flush_interval, self.flush)
+        self._timer.start()
 
     @property
     def flush_count(self):
--- a/google/cloud/bigtable/batcher.py
+++ b/google/cloud/bigtable/batcher.py
@@ -373,7 +373,6 @@ class MutationsBatcher(object):
         :raises:
             * :exc:`.batcherMutationsBatchError` if there's any error in the mutations.
         """
-        self._is_open = False
         self.flush()
         self._executor.shutdown(wait=True)
         atexit.unregister(self.close)
--- a/google/cloud/bigtable/batcher.py
+++ b/google/cloud/bigtable/batcher.py
@@ -18,10 +18,10 @@ import queue
 import concurrent.futures
 import atexit
 
-FLUSH_COUNT = 1000
+FLUSH_COUNT = 100
 MAX_MUTATIONS = 100000
-MAX_ROW_BYTES = 5242880  # 5MB
-
+MAX_ROW_BYTES = 20 * 1024 * 1024  # 20MB
+MAX_BATCH_SIZE = 100 * 1024 * 1024
 
 class MaxMutationsError(ValueError):
     """The number of mutations for bulk request is too big."""
@@ -60,10 +60,12 @@ class _MutationsBatchQueue(object):
                     item.row_key, mutation_count
                 )
             )
-        self.total_size += item.get_mutations_size()
-        self.total_mutation_count += mutation_count
+
 
         self._queue.put(item, block=block, timeout=timeout)
+
+        self.total_size += item.get_mutations_size()
+        self.total_mutation_count += mutation_count
 
     def full(self):
         """Check if the queue is full."""
--- a/google/cloud/bigtable/batcher.py
+++ b/google/cloud/bigtable/batcher.py
@@ -49,7 +49,7 @@ class _MutationsBatchQueue(object):
         self._queue = queue.Queue()
         self.total_mutation_count = 0
         self.total_size = 0
-        self.max_row_bytes = max_row_bytes
+        self.max_mutation_bytes = max_mutation_bytes
         self.flush_count = flush_count
 
     def get(self):
@@ -74,7 +74,7 @@ class _MutationsBatchQueue(object):
         """Check if the queue is full."""
         if (
             self.total_mutation_count >= self.flush_count
-            or self.total_size >= self.max_row_bytes
+            or self.total_size >= self.max_mutation_bytes
         ):
             return True
         return False
@@ -94,7 +94,9 @@ class BatchInfo:
 
 class FlowControl(object):
     def __init__(
-        self, max_mutations=MAX_OUTSTANDING_ELEMENTS, max_row_bytes=MAX_OUTSTANDING_BYTES
+        self,
+        max_mutations=MAX_OUTSTANDING_ELEMENTS,
+        max_mutation_bytes=MAX_OUTSTANDING_BYTES,
     ):
         """Control the inflight requests. Keep track of the mutations, row bytes and row counts.
         As requests to backend are being made, adjust the number of mutations being processed.
@@ -103,22 +105,21 @@ class FlowControl(object):
         Reopen the flow as requests are finished.
         """
         self.max_mutations = max_mutations
-        self.max_row_bytes = max_row_bytes
+        self.max_mutation_bytes = max_mutation_bytes
         self.inflight_mutations = 0
         self.inflight_size = 0
-        self.inflight_rows_count = 0  # TODO: FOR DEBUG, DELETE before merging
         self.event = threading.Event()
         self.event.set()
 
     def is_blocked(self):
         """Returns True if:
         - inflight mutations >= max_mutations, or
-        - inflight bytes size >= max_row_bytes, or
+        - inflight bytes size >= max_mutation_bytes, or
         """
 
         return (
             self.inflight_mutations >= self.max_mutations
-            or self.inflight_size >= self.max_row_bytes
+            or self.inflight_size >= self.max_mutation_bytes
         )
 
     def control_flow(self, batch_info):
@@ -128,8 +129,6 @@ class FlowControl(object):
 
         self.inflight_mutations += batch_info.mutations_count
         self.inflight_size += batch_info.mutations_size
-        self.inflight_rows_count += batch_info.rows_count
-
         self.set_flow_control_status()
         self.wait()
 
@@ -157,7 +156,6 @@ class FlowControl(object):
         """
         self.inflight_mutations -= batch_info.mutations_count
         self.inflight_size -= batch_info.mutations_size
-        self.inflight_rows_count -= batch_info.rows_count
         self.set_flow_control_status()
 
 
@@ -204,7 +202,7 @@ class MutationsBatcher(object):
         flush_interval=1,
     ):
         self._rows = _MutationsBatchQueue(
-            max_row_bytes=max_row_bytes, flush_count=flush_count
+            max_mutation_bytes=max_row_bytes, flush_count=flush_count
         )
         self.table = table
         self._executor = concurrent.futures.ThreadPoolExecutor()
@@ -213,7 +211,7 @@ class MutationsBatcher(object):
         self._timer.start()
         self.flow_control = FlowControl(
             max_mutations=MAX_OUTSTANDING_ELEMENTS,
-            max_row_bytes=MAX_OUTSTANDING_BYTES,
+            max_mutation_bytes=MAX_OUTSTANDING_BYTES,
         )
         self.futures_mapping = {}
         self.exceptions = queue.Queue()
@@ -224,7 +222,7 @@ class MutationsBatcher(object):
 
     @property
     def max_row_bytes(self):
-        return self._rows.max_row_bytes
+        return self._rows.max_mutation_bytes
 
     def __enter__(self):
         """Starting the MutationsBatcher as a context manager"""
@@ -325,7 +323,7 @@ class MutationsBatcher(object):
                 rows_count >= self.flush_count
                 or mutations_size >= self.max_row_bytes
                 or mutations_count >= self.flow_control.max_mutations
-                or mutations_size >= self.flow_control.max_row_bytes
+                or mutations_size >= self.flow_control.max_mutation_bytes
                 or self._rows.empty()  # submit when it reached the end of the queue
             ):
                 self.flow_control.control_flow(batch_info)
--- a/google/cloud/bigtable/batcher.py
+++ b/google/cloud/bigtable/batcher.py
@@ -129,7 +129,7 @@ class MutationsBatcher(object):
 
     def mutate(self, row):
         """Add a row to the batch. If the current batch meets one of the size
-        limits, the batch is sent synchronously.
+        limits, the batch is sent asynchronously.
 
         For example:
 
@@ -156,7 +156,7 @@ class MutationsBatcher(object):
 
     def mutate_rows(self, rows):
         """Add multiple rows to the batch. If the current batch meets one of the size
-        limits, the batch is sent synchronously.
+        limits, the batch is sent asynchronously.
 
         For example:
 
@@ -180,7 +180,7 @@ class MutationsBatcher(object):
             self.mutate(row)
 
     def flush(self):
-        """Sends the current. batch to Cloud Bigtable.
+        """Sends the current batch to Cloud Bigtable synchronously.
         For example:
 
         .. literalinclude:: snippets.py
@@ -195,6 +195,8 @@ class MutationsBatcher(object):
         self.flush_rows(rows_to_flush)
 
     def flush_async(self):
+        """Sends the current batch to Cloud Bigtable asynchronously."""
+
         rows_to_flush = []
         while not self._rows.empty():
             rows_to_flush.append(self._rows.get())
--- a/google/cloud/bigtable/batcher.py
+++ b/google/cloud/bigtable/batcher.py
@@ -94,7 +94,7 @@ class BatchInfo:
 
 class FlowControl(object):
     def __init__(
-        self, max_mutations=MAX_OUTSTANDING_BYTES, max_row_bytes=MAX_MUTATION_SIZE
+        self, max_mutations=MAX_OUTSTANDING_ELEMENTS, max_row_bytes=MAX_OUTSTANDING_BYTES
     ):
         """Control the inflight requests. Keep track of the mutations, row bytes and row counts.
         As requests to backend are being made, adjust the number of mutations being processed.
--- a/google/cloud/bigtable/batcher.py
+++ b/google/cloud/bigtable/batcher.py
@@ -25,9 +25,9 @@ from dataclasses import dataclass
 
 FLUSH_COUNT = 100  # after this many elements, send out the batch
 
-MAX_ROW_BYTES = 20 * 1024 * 1024  # 20MB # after this many bytes, send out the batch
+MAX_MUTATION_SIZE = 20 * 1024 * 1024  # 20MB # after this many bytes, send out the batch
 
-MAX_MUTATIONS_SIZE = 100 * 1024 * 1024  # 100MB # max inflight byte size.
+MAX_OUTSTANDING_BYTES = 100 * 1024 * 1024  # 100MB # max inflight byte size.
 
 MAX_OUTSTANDING_ELEMENTS = 100000  # max inflight mutations.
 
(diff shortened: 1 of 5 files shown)


google/cloud/bigtable/batcher.py
```diff
@@ -24,6 +27,74 @@ class MaxMutationsError(ValueError):
...
+        """Insert an item to the queue. Recalculate queue size."""
+
+        mutation_count = len(item._get_mutations())
+
+        if mutation_count > MAX_MUTATIONS:
+            raise MaxMutationsError(
+                "The row key {} exceeds the number of mutations {}.".format(
+                    item.row_key, mutation_count
+                )
+            )
```
mutianf, line 62: We should also block when the total inflight size is > 100MB. I think we'll need another variable FLUSH_SIZE to track when the current batch is full, and track the total size separately.

google/cloud/bigtable/batcher.py
```diff
@@ -139,8 +228,30 @@ def flush(self):
...
+            self.table.mutate_rows(rows_to_flush)
+
+    def __exit__(self, exc_type, exc_value, exc_traceback):
+        """Clean up resources. Flush and shutdown the ThreadPoolExecutor."""
+        self.close()
+
+    def close(self):
+        """Clean up resources. Flush and shutdown the ThreadPoolExecutor."""
+        self._is_open = False
+        self.flush()
```
mutianf, line 256: how are errors handled? the java client throw errors on the failed elements onClose(). Will the batcher fail silently if some elements failed to apply?

google/cloud/bigtable/batcher.py
```diff
@@ -24,6 +27,74 @@ class MaxMutationsError(ValueError):
...
 
 
+class BatcherIsClosedError(ValueError):
+    """Batcher is already closed and not accepting any mutation."""
+
+
+class _MutationsBatchQueue(object):
+    """Private Threadsafe Queue to hold rows for batching."""
+
+    def __init__(self, max_row_bytes=MAX_ROW_BYTES, flush_count=FLUSH_COUNT):
```
mutianf, line 37: as a user, how can i reconfigure max bytes and flush count? (This can come in a later pr)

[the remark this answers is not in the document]
Mariatta: Sure, I will update the numbers to match the ones set for Java.

Mariatta: These values gets passed from the `MutationsBatcher` constructor.

For example, they can do:
``` 
batcher = MutationsBatcher(max_row_bytes=5*1024, flush_count=50)
```

google/cloud/bigtable/batcher.py
Mariatta: It uses the existing retry mechanism in `table.mutate_rows`.

https://github.com/googleapis/python-bigtable/blob/main/google/cloud/bigtable/table.py#L680

Mariatta: So `table.mutate_rows` returns the response statuses to indicate errors.
We can capture the status codes and raise error.

google/cloud/bigtable/batcher.py
```diff
@@ -24,6 +27,74 @@ class MaxMutationsError(ValueError):
...
+        self.total_mutation_count -= len(row._get_mutations())
+        self.total_size -= mutation_size
+        return row
+
+    def put(self, item, block=True, timeout=None):
+        """Insert an item to the queue. Recalculate queue size."""
+
+        mutation_count = len(item._get_mutations())
+
+        if mutation_count > MAX_MUTATIONS:
```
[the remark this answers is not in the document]
Mariatta: This is the max number of mutations for a single row. (100,000) In this case I think it should raise an error since we shouldn't split the mutations for one row?

google/cloud/bigtable/batcher.py
```diff
@@ -13,17 +13,108 @@
...
+import concurrent.futures
+import atexit
 
 
-FLUSH_COUNT = 1000
+FLUSH_COUNT = 100
 MAX_MUTATIONS = 100000
-MAX_ROW_BYTES = 5242880  # 5MB
+MAX_ROW_BYTES = 20 * 1024 * 1024  # 20MB
+MAX_MUTATIONS_SIZE = 100 * 1024 * 1024  # 100MB
```
igorbernstein2, line 25: Please add comments what these constants control

google/cloud/bigtable/batcher.py
```diff
@@ -13,17 +13,108 @@
...
+    """Private Threadsafe Queue to hold rows for batching."""
+
+    def __init__(self, max_row_bytes=MAX_ROW_BYTES, flush_count=FLUSH_COUNT):
+        """Specify the queue constraints"""
+        self._queue = queue.Queue(maxsize=flush_count)
+        self.total_mutation_count = 0
+        self.total_size = 0
+        self.max_row_bytes = max_row_bytes
+
+    def get(self, block=True, timeout=None):
```
igorbernstein2, line 55: when would block = False?

google/cloud/bigtable/batcher.py
```diff
@@ -54,19 +145,49 @@ class MutationsBatcher(object):
     flush. If it reaches the max number of row mutations size it calls
     finish_batch() to mutate the current row batch. Default is MAX_ROW_BYTES
     (5 MB).
+
```
igorbernstein2, line 148: Please remove the TODO from the doc
Also please document the threadsafety behavior. From what I see in the code, a Batcher can't be shared by multiple enduser  threads?

google/cloud/bigtable/batcher.py
```diff
@@ -125,22 +236,86 @@ def mutate_rows(self, rows):
...
+        """
+
+        rows_to_flush = []
+        while not self._rows.empty():
+            rows_to_flush.append(self._rows.get())
+        future = self._executor.submit(self.flush_rows, rows_to_flush)
+        # catch the exceptions in the mutation
+        exc = future.exception()
+        if exc:
+            raise exc
```
igorbernstein2, line 279: will this throw the original exception or will it wrap it?
If its the original exception, then the stacktrace will be rooted in the executor. If its wrapped then you are leaking implementation details.

In java we kept the original exception but added a synthetic stacktrace of the caller to help callers diagnose where they called the failed RPC.

HBase does something even hackier and modifies the stacktrace and inserts the callers stack at the top.

I dont know which approach idiomatic to python, but we should be intentional here

igorbernstein2 (changes requested):
It seems like we don't really have any flow control here. Do you plan on adding it in a follow up PR?

google/cloud/bigtable/batcher.py
Mariatta: This will re-raise the original exception. I will see how to do that in Python.
Can you share link on how it was done in Java?

google/cloud/bigtable/batcher.py
```diff
@@ -125,22 +236,86 @@ def mutate_rows(self, rows):
...
+            response = self.table.mutate_rows(rows_to_flush)
+            responses = []
+            has_error = False
+            for result in response:
+                if result.code != 0:
+                    has_error = True
+                responses.append(result)
+
+            if has_error:
+                raise MutationsBatchError(status_codes=response)
```
daniel-sanche, line 303: I noticed Status codes can be converted into exceptions using [google.api_core.exceptions.from_grpc_status](https://github.com/googleapis/python-api-core/blob/b6eb6dee2762b602d553ace510808afd67af461d/google/api_core/exceptions.py#L441). Maybe it would be better to raise those directly, to give the full context, rather than using the raw status codes?

google/cloud/bigtable/batcher.py
Mariatta: In this case it the exception will have the stack trace all the way to the original spot where it was raised (line 303 below). The exception itself will have a list of individual error codes so user can still go through it to find out what failed.

google/cloud/bigtable/batcher.py
Mariatta: I've removed the TODO. Correct the same batcher cannot be shared by multiple enduser threads.

google/cloud/bigtable/batcher.py
Mariatta: An error is now raised if any of the mutation errored out.

google/cloud/bigtable/batcher.py
```diff
@@ -13,17 +13,121 @@
...
+
+# Max size (in bytes) for a single mutation
+MAX_ROW_BYTES = 20 * 1024 * 1024  # 20MB
+
+# Max size (in bytes) for a single request
+MAX_MUTATIONS_SIZE = 100 * 1024 * 1024  # 100MB
+
+
+class MutationsBatchError(Exception):
+    """Error in the batch request"""
```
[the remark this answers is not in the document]
Mariatta: Yes I've thought of the exception group and as you know it's only in Python 3.11. Your suggestion sounds great, I'll make the changes.

google/cloud/bigtable/batcher.py
```diff
@@ -13,17 +13,117 @@
...
+
+        self.total_size += item.get_mutations_size()
+        self.total_mutation_count += mutation_count
+
+    def full(self):
+        """Check if the queue is full."""
+        if (
+            self.total_mutation_count >= MAX_MUTATIONS
+            or self.total_size >= self.max_row_bytes
+            or self._queue.full()
```
mutianf, line 102: Why do we need to check if the queue is full? In the batcher code, we flush when the Batcher is full:
```
if self._rows.full():
    self.flush_async()
```
I'm not sure if the behavior is still correct? 

I think full() should only check for 2 things: 
- Number of elements reached the batch element threshold
- Number of bytes reached the batch bytes threshold

The queue is only be used to block user from adding more elements when it's full. We don't trigger another flush when the queue is full.

google/cloud/bigtable/batcher.py
```diff
@@ -125,22 +244,92 @@ def mutate_rows(self, rows):
...
             :end-before: [END bigtable_api_batcher_flush]
             :dedent: 4
 
+        raises:
+            * :exc:`.batcherMutationsBatchError` if there's any error in the
+                mutations.
+        """
+        rows_to_flush = []
+        while not self._rows.empty():
+            rows_to_flush.append(self._rows.get())
```
mutianf, line 268: would this also remove the row from the queue? In this case, we can't block on adding more elements to the queue, or when we block it doesn't necessarily mean that we've reached the max outstanding bytes or elements?

google/cloud/bigtable/batcher.py
```diff
@@ -125,22 +244,92 @@ def mutate_rows(self, rows):
...
+                mutations.
+        """
+
+        rows_to_flush = []
+        while not self._rows.empty():
+            rows_to_flush.append(self._rows.get())
+        future = self._executor.submit(self.flush_rows, rows_to_flush)
+        # catch the exceptions in the mutation
+        exc = future.exception()
+        if exc:
```
mutianf, line 286: I think we may want to only raise the errors when we close the batcher

google/cloud/bigtable/batcher.py
```diff
@@ -125,22 +298,120 @@ def mutate_rows(self, rows):
...
+                future.add_done_callback(self.batch_completed_callback)
+
+                # reset and start a new batch
+                rows_to_flush = []
+                mutations_size = 0
+                rows_count = 0
+                mutations_count = 0
+                batch_info = BatchInfo()
+
+    def batch_completed_callback(self, future):
```
daniel-sanche, line 369: is this meant to be public?

google/cloud/bigtable/batcher.py
```diff
@@ -125,22 +298,120 @@ def mutate_rows(self, rows):
...
-            self.table.mutate_rows(self.rows)
-            self.total_mutation_count = 0
-            self.total_size = 0
-            self.rows = []
+        self._is_open = False
+        self.flush()
+        self._executor.shutdown(wait=True)
+        if self.exceptions.qsize() > 0:
+            exc = list(self.exceptions.queue)
+            raise MutationsBatchError("Errors in batch mutations.", exc=exc)
```
daniel-sanche, line 417: Maybe also unregister the `atexit` callback registered in `__init__`?

google/cloud/bigtable/batcher.py
```diff
@@ -125,22 +298,120 @@ def mutate_rows(self, rows):
...
+        raises:
+            * :exc:`.batcherMutationsBatchError` if there's any error in the
+                mutations.
+        """
+
+        rows_to_flush = []
+        mutations_count = 0
+        mutations_size = 0
+        rows_count = 0
+        batch_info = BatchInfo()
```
daniel-sanche, line 338: I'm a bit confused by `BatchInfo`. Doesn't it duplicate the `row_count`, `mutations_count` and `mutations_size` variables?

google/cloud/bigtable/batcher.py
```diff
@@ -125,22 +298,120 @@ def mutate_rows(self, rows):
...
+
+        Raise exceptions if there's any.
+        Release the resources locked by the flow control and allow enqueued tasks to be run.
+        """
+
+        processed_rows = self.futures_mapping[future]
+        self.flow_control.release(processed_rows)
+        del self.futures_mapping[future]
+
+    def flush_rows(self, rows_to_flush=None):
```
daniel-sanche, line 380: Having both `flush()` and `flush_rows()` seems a little confusing. Especially since this doesn't seem to be "flushing" from the cache in any way. I could see people calling this with no arguments thinking it is the main flush function

Maybe this should be called `mutate_rows`? Or just made into an internal helper function?

google/cloud/bigtable/batcher.py
```diff
@@ -125,22 +298,120 @@ def mutate_rows(self, rows):
...
+        """Mutate the specified rows.
+
+        raises:
+            * :exc:`.batcherMutationsBatchError` if there's any error in the
+                mutations.
+        """
+        responses = []
+        if len(rows_to_flush) > 0:
```
daniel-sanche, line 388: this would raise a `TypeError` if `flush_rows` is called without any arguments. Is that the intention?

google/cloud/bigtable/batcher.py
```diff
@@ -125,22 +298,120 @@ def mutate_rows(self, rows):
...
+            rows_count += 1
+            rows_to_flush.append(row)
+            batch_info.mutations_count = mutations_count
+            batch_info.rows_count = rows_count
+            batch_info.mutations_size = mutations_size
+
+            if (
+                rows_count >= self.flush_count
+                or mutations_size >= self.max_row_bytes
+                or mutations_count >= self.flow_control.max_mutations
```
mutianf, line 353: I dont think we need to check flow control limits here since you're calling `self.flow_control.control_flow(batch_info)` on line 357.

google/cloud/bigtable/batcher.py
```diff
@@ -125,22 +298,120 @@ def mutate_rows(self, rows):
...
+                or mutations_size >= self.flow_control.max_row_bytes
+                or self._rows.empty()  # submit when it reached the end of the queue
```
mutianf, line 355: why not call _rows.isfull()?

google/cloud/bigtable/batcher.py
Mariatta: Changed to private.

google/cloud/bigtable/batcher.py
Mariatta: It was a mistake that the `rows_to_flush` was optional. I've changed it to required.

google/cloud/bigtable/batcher.py
Mariatta: Yes, added.

google/cloud/bigtable/batcher.py
Mariatta: Agree this is confusing. There is already a public `mutate_rows` from before, and it has different behavior than this one. I'm changing it to private since user isn't expected to call it manually.

google/cloud/bigtable/batcher.py
Mariatta: In here, we're popping an elements from `self._rows` (line 341), taking items out from the queue, perform mutations in a batch, and waiting for response.
We can't call `_rows.full()` because it is being emptied at this point, so it doesn't become full.

google/cloud/bigtable/batcher.py
Mariatta: We still need to check it here. At this point the flow control has not been activated yet. We're checking if we will need to activate the flow control after this condition is met. Without this check it will not actually activate the flow control.

google/cloud/bigtable/batcher.py
Mariatta: It's similar info, but for different "buckets".

There are two "queues":

`self._rows`: for storing the rows we want to mutate
`batch_info`: for storing info about the rows that are being mutated, and we're waiting for the result/response from backend. This gets passed to the the batch_completed_callback. That's where we can release the flow control.

google/cloud/bigtable/batcher.py
Mariatta: Yes rows gets popped out of the queue. I have added flow control mechanism after your comment about this. So this is outdated.

google/cloud/bigtable/batcher.py
```diff
@@ -13,17 +13,172 @@
...
+        )
+
+    def control_flow(self, batch_info):
+        """
+        Calculate the resources used by this batch
+        """
+
+        self.inflight_mutations += batch_info.mutations_count
+        self.inflight_size += batch_info.mutations_size
+        self.inflight_rows_count += batch_info.rows_count
```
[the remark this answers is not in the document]
Mariatta: I was using this for debugging. I will remove it in the end before merging.

google/cloud/bigtable/batcher.py
```diff
@@ -13,17 +13,172 @@
...
+        return row
+
+    def put(self, item):
+        """Insert an item to the queue. Recalculate queue size."""
+
+        mutation_count = len(item._get_mutations())
+        mutation_size = item.get_mutations_size()
+
+        if mutation_count > MAX_OUTSTANDING_ELEMENTS:
+            raise MaxMutationsError(
```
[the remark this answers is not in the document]
Mariatta: Since queue.put (line 87) is blocking and waiting until it becomes free again, I'm removing the errors.

google/cloud/bigtable/batcher.py
```diff
@@ -13,17 +13,172 @@
...
+                "The row key {} exceeds the number of mutations {}.".format(
+                    item.row_key, mutation_count
+                )
+            )
+
+        if mutation_size > MAX_MUTATIONS_SIZE:
```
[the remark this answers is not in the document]
Mariatta: queue.put() is already blocking and waiting until it has space available. So I'm removing this error.

google/cloud/bigtable/batcher.py
```diff
@@ -13,17 +13,172 @@
...
+class BatchInfo:
+    """Keeping track of size of a batch"""
+
+    mutations_count: int = 0
+    rows_count: int = 0
+    mutations_size: int = 0
+
+
+class FlowControl(object):
+    def __init__(self, max_mutations=MAX_MUTATIONS_SIZE, max_row_bytes=MAX_ROW_BYTES):
```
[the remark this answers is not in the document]
Mariatta: And these numbers are:
MAX_OUTSTANDING_ELEMENTS  = 100000
MAX_MUTATION_SIZE = 20 MB 
correct?

google/cloud/bigtable/batcher.py
Mariatta: This has been addressed.

google/cloud/bigtable/batcher.py
```diff
@@ -13,15 +13,152 @@
...
+    def __init__(self, message, exc):
+        self.exc = exc
+        self.message = message
+        super().__init__(self.message)
+
+
+class _MutationsBatchQueue(object):
+    """Private Threadsafe Queue to hold rows for batching."""
+
+    def __init__(self, max_row_bytes=MAX_MUTATION_SIZE, flush_count=FLUSH_COUNT):
```
mutianf, line 47: nit:
```suggestion
    def __init__(self, max_mutation_bytes=MAX_MUTATION_SIZE, flush_count=FLUSH_COUNT):
```

google/cloud/bigtable/batcher.py
```diff
@@ -13,15 +13,152 @@
...
+    """Keeping track of size of a batch"""
+
+    mutations_count: int = 0
+    rows_count: int = 0
+    mutations_size: int = 0
+
+
+class FlowControl(object):
+    def __init__(
+        self, max_mutations=MAX_OUTSTANDING_BYTES, max_row_bytes=MAX_MUTATION_SIZE
```
mutianf, line 97: defaults should be:
```suggestion
        self, max_mutations=MAX_OUTSTANDING_ELEMENTS, max_row_bytes=MAX_OUTSTANDING_BYTES
```

mutianf (approved):
LGTM with some nits!

daniel-sanche (approved):

google/cloud/bigtable/batcher.py
```diff
@@ -13,15 +13,152 @@
...
+        As requests to backend are being made, adjust the number of mutations being processed.
+
+        If threshold is reached, block the flow.
+        Reopen the flow as requests are finished.
+        """
+        self.max_mutations = max_mutations
+        self.max_row_bytes = max_row_bytes
+        self.inflight_mutations = 0
+        self.inflight_size = 0
+        self.inflight_rows_count = 0  # TODO: FOR DEBUG, DELETE before merging
```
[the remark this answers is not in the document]
Mariatta: Deleted

igorbernstein2 (changes requested):

google/cloud/bigtable/batcher.py
```diff
@@ -13,104 +13,252 @@
...
+@dataclass
+class _BatchInfo:
+    """Keeping track of size of a batch"""
+
+    mutations_count: int = 0
+    rows_count: int = 0
+    mutations_size: int = 0
+
+
+class FlowControl(object):
```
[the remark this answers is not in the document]
Mariatta: Sure, I will add that in the next commit.

google/cloud/bigtable/batcher.py
```diff
@@ -13,104 +13,252 @@
...
+
+        If threshold is reached, block the flow.
+        Reopen the flow as requests are finished.
+        """
+        self.max_mutations = max_mutations
+        self.max_mutation_bytes = max_mutation_bytes
+        self.inflight_mutations = 0
+        self.inflight_size = 0
+        self.event = threading.Event()
+        self.event.set()
```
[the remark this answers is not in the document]
daniel-sanche: Since I also was looking at something similar for the v3 branch: Python semaphores only allow incrementing by 1 at a time, but we'd need to increment it by the thousands to track bytes.

My current solution was to use a [condition](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Condition), with a check against the flow limits as a predicate. But I think the event can be set up in a similar way

google/cloud/bigtable/batcher.py
```diff
@@ -13,104 +13,252 @@
...
+
+    def control_flow(self, batch_info):
+        """
+        Calculate the resources used by this batch
+        """
+
+        self.inflight_mutations += batch_info.mutations_count
+        self.inflight_size += batch_info.mutations_size
+        self.set_flow_control_status()
+        self.wait()
```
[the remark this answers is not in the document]
Mariatta: I adjusted the logic, moving the `wait` to the `flush_async` function. If the batch is causing the event to be blocked, then it will be sent through, but the subsequent thread will be blocked and waited.

google/cloud/bigtable/batcher.py
```diff
@@ -119,28 +267,115 @@ def mutate_rows(self, rows):
...
             :end-before: [END bigtable_api_batcher_flush]
             :dedent: 4
 
+        :raises:
+            * :exc:`.batcherMutationsBatchError` if there's any error in the mutations.
+        """
+        rows_to_flush = []
+        while not self._rows.empty():
+            rows_to_flush.append(self._rows.get())
+        response = self._flush_rows(rows_to_flush)
```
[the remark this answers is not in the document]
Mariatta: This function is executed per second, and when the batch gets closed, when we're supposed to send all remaining batches sequentially.

igorbernstein2 (approved):
LGTM, but please have Daniel take a look at the api and confirm it works for him in the async client

daniel-sanche:
Some of the API will need to change for the async client, due to different model classes and asyncio patterns, but the general shape of the solution should be mostly consistent. LGTM

PR #751, opened 2023-03-16
feat: improve rows filters

daniel-sanche:
This PR builds off of the existing RowFilters classes, witha  couple additions:
- adds `to_dict` methods, making it easier to work with dict representations of the objects
- adds type annotations
- adds `str` and `repr` implementations to make filters more human readable
- _FilterCombination types are Sequences (implement `len`, iteration, indexing, etc)

I also pulled in the existing unit test file into the new client's test folder, with added tests for the new functionality

TODO:
- changes docstrings to be consistent with other classes
---

### DIFF
--- a/google/cloud/bigtable/row_filters.py
+++ b/google/cloud/bigtable/row_filters.py
@@ -17,7 +17,7 @@ from __future__ import annotations
 
 import struct
 
-from typing import Any, TYPE_CHECKING
+from typing import Any, Sequence, TYPE_CHECKING, overload
 
 from google.cloud._helpers import _microseconds_from_datetime  # type: ignore
 from google.cloud._helpers import _to_bytes  # type: ignore
@@ -758,7 +758,17 @@ class _FilterCombination(RowFilter, Sequence[RowFilter]):
     def __len__(self) -> int:
         return len(self.filters)
 
+    @overload
     def __getitem__(self, index: int) -> RowFilter:
+        # overload signature for type checking
+        pass
+
+    @overload
+    def __getitem__(self, index: slice) -> list[RowFilter]:
+        # overload signature for type checking
+        pass
+
+    def __getitem__(self, index):
         return self.filters[index]
 
     def __repr__(self) -> str:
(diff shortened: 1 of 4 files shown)


google/cloud/bigtable/row_filters.py
```diff
@@ -35,6 +41,19 @@ class RowFilter(object):
         This class is a do-nothing base class for all row filters.
     """
 
+    def to_pb(self) -> data_v2_pb2.RowFilter:
```
daniel-sanche, line 44: The original implementation of RowFilter had `to_pb` functions for each filter, but I don't think we'll need it for the v3 client. I'm leaning towards removing `to_pb` for simplicity (and in case we change proto formats in the future), but I'm curious if there are any reasons for keeping it around, since it exists and is well tested.

google/cloud/bigtable/row_filters.py
```diff
@@ -35,6 +41,21 @@ class RowFilter(object):
...
+        """Converts the row filter to a protobuf.
+
+        Returns: The converted current object.
+        """
+        return data_v2_pb2.RowFilter(**self.to_dict())
+
+    def to_dict(self) -> dict[str, Any]:
+        """Converts the row filter to a dict representation."""
+        # unimplemented on base class
+        raise NotImplementedError
```
Mariatta, line 54: What do you think of creating an ABC using abc.ABC for `RowFilter` so that `to_dict` can be one of its `@abstractmethod`

tests/unit/test_row_filters.py
```diff
@@ -0,0 +1,2000 @@
+# Copyright 2016 Google LLC
```
Mariatta, line 1:
```suggestion
# Copyright 2023 Google LLC
```

google/cloud/bigtable/row_filters.py
daniel-sanche: Yeah I think that makes sense. The only complication is that it requires us to make other filter subclasses abstract as well (_BoolFilter, _FilterCombination, _CellCountFilter), but that's probably good to prevent users from instantiating them directly anyway

I made the changes, let me know what you think

google/cloud/bigtable/row_filters.py
Mariatta: I would suggest keeping it public, because it is already available in v2. We don't know whether users are relying on this or not. I think it will make it easier for the to upgrade to v3 if they have access to similar functions. Otherwise we need to start raising deprecation warning saying this will be removed in v3.

Mariatta (approved):

PR #752, opened 2023-03-16
feat: read rows query model class

daniel-sanche:
Implements the RowQuery class, used to configure read_rows calls

Implementation is generally aligned with the [Query class in the Java library](https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java)

Implementation for sharding will come in a future PR

---

### DIFF
--- a/google/cloud/bigtable/read_rows_query.py
+++ b/google/cloud/bigtable/read_rows_query.py
@@ -76,7 +76,7 @@ class ReadRowsQuery:
 
     def __init__(
         self,
-        row_keys: list[str | bytes] | str | bytes | None = None,
+        row_key: list[str | bytes] | str | bytes | None = None,
         row_ranges: list[RowRange] | RowRange | None = None,
         limit: int | None = None,
         row_filter: RowFilter | dict[str, Any] | None = None,
@@ -85,7 +85,9 @@ class ReadRowsQuery:
         Create a new ReadRowsQuery
 
         Args:
-          - row_keys: a list of row keys to include in the query
+          - row_keys: row keys to include in the query
+                a query can contain multiple keys, but ranges should be preferred
+          - row_ranges: ranges of rows to include in the query
           - limit: the maximum number of rows to return. None or 0 means no limit
                 default: None (no limit)
           - row_filter: a RowFilter to apply to the query
@@ -95,7 +97,8 @@ class ReadRowsQuery:
         for range in row_ranges:
             self.row_ranges.append(range)
         if row_keys:
-            self.add_rows(row_keys)
+            for k in row_keys:
+                self.add_key(k)
         self.limit: int | None = limit
         self.filter: RowFilter | dict[str, Any] = row_filter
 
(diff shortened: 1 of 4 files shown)


tests/unit/test_read_rows_query.py
```diff
@@ -0,0 +1,284 @@
...
+        query = self._make_one(["row_key_1", "row_key_2"], limit=10, row_filter=filter_)
+        self.assertEqual(len(query.row_keys), 2)
+        self.assertIn("row_key_1".encode(), query.row_keys)
+        self.assertIn("row_key_2".encode(), query.row_keys)
+        self.assertEqual(query.row_ranges, [])
+        self.assertEqual(query.filter, filter_)
+        self.assertEqual(query.limit, 10)
+
+    def test_ctor_invalid_limit(self):
+        with self.assertRaises(ValueError):
```
Mariatta, line 53: Can you also assert that the valueerror message == "limit must be >= 0"?

tests/unit/test_read_rows_query.py
```diff
@@ -0,0 +1,284 @@
...
+        self.assertEqual(query.row_ranges[3][1].key, b"test_row7")
+        self.assertEqual(query.row_ranges[3][0].is_inclusive, False)
+        self.assertEqual(query.row_ranges[3][1].is_inclusive, True)
+        # test with nothing passed
+        result = query.add_range()
+        self.assertEqual(len(query.row_ranges), 5)
+        self.assertEqual(query.row_ranges[4][0], None)
+        self.assertEqual(query.row_ranges[4][1], None)
+        # test with inclusive flags only
+        with self.assertRaises(ValueError):
```
Mariatta, line 215: `add_range` seems to be raising ValueError with varying messages. Perhaps we can also check which error message is being raised in each of the errors below.

daniel-sanche: good point, I added that to all the raises statements

google/cloud/bigtable/read_rows_query.py
```diff
@@ -13,36 +13,158 @@
...
     """
     Class to encapsulate details of a read row request
     """
 
     def __init__(
-        self, row_keys: list[str | bytes] | str | bytes | None = None, limit=None
+        self,
+        row_keys: list[str | bytes] | str | bytes | None = None,
+        limit: int | None = None,
+        row_filter: RowFilter | dict[str, Any] | None = None,
```
mutianf, line 41: __init__ doesn't have row_ranges. is this intended?

google/cloud/bigtable/read_rows_query.py
```diff
@@ -13,36 +13,158 @@
...
+          - start_is_inclusive: if True, the start key is included in the range
+              defaults to True if None. Must not be included if start_key is None
+          - end_is_inclusive: if True, the end key is included in the range
+              defaults to False if None. Must not be included if end_key is None
+        """
+        # check for invalid combinations of arguments
+        if start_is_inclusive is None:
+            start_is_inclusive = True
+        elif start_key is None:
+            raise ValueError("start_is_inclusive must not be set without start_key")
```
mutianf, line 145: Maybe rephrase it to 
```suggestion
            raise ValueError("start_is_inclusive must be set with start_key")
```
so it's easier to understand :)

google/cloud/bigtable/read_rows_query.py
daniel-sanche: Yes, because this implementation doesn't have an explicit range object, so passing in multiple ranges at init time would be difficult. Users can call `Query().add_range(start, end)` to add ranges

I think Python's style is biased towards flattness, and avoiding custom objects when you can get away with it. But I'm not opposed to adding a range class if you expect users to be re-using or passing ranges around a lot

google/cloud/bigtable/read_rows_query.py
```diff
@@ -13,36 +13,158 @@
...
+from google.cloud.bigtable.row_filters import RowFilter
 
 if TYPE_CHECKING:
-    from google.cloud.bigtable.row_filters import RowFilter
     from google.cloud.bigtable import RowKeySamples
 
 
+@dataclass
+class _RangePoint:
+    # model class for a point in a row range
```
Mariatta, line 27:
```suggestion
    """Model class for a point in a row range"""
```

Mariatta (commented):
One small change, but otherwise LGTM

Mariatta (approved):

PR #753, opened 2023-03-16
feat: implement row and cell model classes

daniel-sanche:
This PR implements the RowResponse and CellResponse classes, used to represent the data returned by read_rows queries

RowResponse implements Sequence, allowing it to be treated like a list of Cells (len, iteration, indexing, etc)

It also implements some properties of Mapping (allowing it to be indexed by `family` and `(family, qualifier)` keys, retrieving the list of keys, values, items, etc)

---

### DIFF
--- a/google/cloud/bigtable/row_response.py
+++ b/google/cloud/bigtable/row_response.py
@@ -15,7 +15,8 @@
 from __future__ import annotations
 
 from collections import OrderedDict
-from typing import Sequence
+from typing import Sequence, Generator, Mapping, overload, Union, List, Tuple, Any
+from functools import total_ordering
 
 # Type aliases used internally for readability.
 row_key = bytes
@@ -24,7 +25,12 @@ qualifier = bytes
 row_value = bytes
 
 
-class RowResponse(Sequence["CellResponse"]):
+class RowResponse(
+    Sequence["CellResponse"],
+    Mapping[
+        Union[family_id, Tuple[family_id, Union[qualifier, str]]], List["CellResponse"]
+    ],
+):
     """
     Model class for row data returned from server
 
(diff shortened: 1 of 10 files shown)


google/cloud/bigtable/row_response.py
```diff
@@ -36,47 +42,237 @@ class RowResponse(Sequence["CellResponse"]):
...
 
     def get_cells(
-        self, family: str | None, qualifer: str | bytes | None
+        self, family: str | None = None, qualifier: str | bytes | None = None
     ) -> list[CellResponse]:
         """
         Returns cells sorted in Bigtable native order:
             - Family lexicographically ascending
-            - Qualifier lexicographically ascending
+            - Qualifier ascending
```
daniel-sanche, line 86: I had in my notes that Qualifier should be sorted lexographically, but I assume this is a mistake, since qualifier is bytes? 
I implemented this by directly comparing the bytes values instead

mutianf: java-bigtable uses a byteStringComaprator: https://github.com/googleapis/java-bigtable/blob/ef7d7419293f9f046dcf941148a86da713feae17/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/ByteStringComparator.java

https://github.com/googleapis/java-bigtable/blob/ef7d7419293f9f046dcf941148a86da713feae17/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/RowCell.java#L51

google/cloud/bigtable/row_response.py
```diff
@@ -91,20 +329,51 @@ def __init__(
...
+        """
+        Helper function to create CellResponse from a dictionary
+
+        CellResponse objects are not intended to be constructed by users.
+        They are returned by the Bigtable backend.
+        """
+        # Bigtable backend will use microseconds for timestamps,
+        # but the Python library prefers nanoseconds where possible
+        timestamp = cell_dict.get(
+            "timestamp_ns", cell_dict.get("timestamp_micros", -1) * 1000
```
mutianf, line 364: what's the benefit of converting it back to nano seconds?

google/cloud/bigtable/row_response.py
```diff
@@ -36,47 +37,284 @@ class RowResponse(Sequence["CellResponse"]):
...
+    def __len__(self):
+        """
+        Implements `len()` operator
+        """
+        return len(self._cells_list)
+
+    def keys(self):
         """
-        raise NotImplementedError
+        Returns a list of (family, qualifier) pairs associated with the cells
```
mutianf, line 260: I would expect keys() to return the row key of the response, maybe there's a better name for this? Or maybe I misunderstand something?

google/cloud/bigtable/row_response.py
```diff
@@ -36,47 +37,284 @@ class RowResponse(Sequence["CellResponse"]):
...
+            )
+        return self._cells_map[family][qualifier]
 
-    def get_index(self) -> dict[family_id, list[qualifier]]:
+    def _get_all_from_family(
+        self, family: family_id
+    ) -> Generator[CellResponse, None, None]:
         """
-        Returns a list of family and qualifiers for the object
+        Returns all cells in the row
```
mutianf, line 120: from the method name maybe you meant
```suggestion
        Returns all cells in the row for the family_id
```

google/cloud/bigtable/row_response.py
```diff
@@ -36,47 +37,284 @@ class RowResponse(Sequence["CellResponse"]):
...
+                    cell_obj = CellResponse._from_dict(
+                        key, family, qualifier, cell_dict
+                    )
+                    tmp_list.append(cell_obj)
+            cells = tmp_list
+        # add cells to internal stores using Bigtable native ordering
+        for cell in sorted(cells):
+            if cell.row_key != self.row_key:
+                raise ValueError(
+                    f"CellResponse row_key ({cell.row_key!r}) does not match RowResponse key ({self.row_key!r})"
```
mutianf, line 71: Is this check necessary?

google/cloud/bigtable/row_response.py
daniel-sanche: Yeah, so for convenience, I made RowResponse conform to the [Sequence base type](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes), so it can be treated as a standard sequence:
- `len(row_response)`
- `for cell in row_response:`,
- `first = row_response[0]`
- `sample = row_response[0:5]`
- etc

I also wanted to make it partially compatible with the [Mapping base type](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes), to make indexing convenient. In this context, the row_response can be treated as a key/value dictionary, where the keys are families or family/qualifier pairs, and the values are the associated cells:
- `cells = row_response["my_family"]`
- `cells = row_response["my-family, "my-qualifier"]`

I added also implementations for Mapping's  [`keys()`](https://docs.python.org/3/library/stdtypes.html#dict.keys) [`values()`](https://docs.python.org/3/library/stdtypes.html#dict.values) and [`items()`](https://docs.python.org/3/library/stdtypes.html#dict.items) functions to make iteration easier:

```
for family,qualifier in row_response.keys():
   cells = row_response[family,qualifier]

for cells in row_response.values():
   print(cells)

for (family,qualifier), cells in row_response.items():
   print(cells)
```

google/cloud/bigtable/row_response.py
daniel-sanche: Python's time module returns time as nanoseconds or (float) seconds, but the Bigtable backend returns time as microseconds. That makes things a bit complicated, since conversion will be required somewhere. I opted to keep the library's canonical representation in nanos, and convert to micros when needed

Since we have these two time representations, I made the `from_dict` function support both, but it still needs to convert it into the canonical representation when building the object

google/cloud/bigtable/row_response.py
daniel-sanche: Our backend should never create rows that hit this, but end users could

And it doesn't hurt to include it as a sanity check either way

google/cloud/bigtable/row_response.py
daniel-sanche: So TL;DR: keys here refers to dictionary.keys(), to make iteration over stored data easier. But I can see how that may be confusing, and I'm ok with making changes here

Mariatta (approved):

google/cloud/bigtable/row_response.py
mutianf: gotcha, thanks for the explanation! _maybe_ we can leave the `implementations for Mapping's keys() values() and items() functions` part out for now? I can see how it's convenient, but maybe we can keep it simple at the first iteration?

google/cloud/bigtable/row_response.py
mutianf: Gotcha, I'm ok with doing the conversion in the implementation, but we should be careful not to expose nano precision on the public surface because a customer could try set the timestamp with nano seconds precision.

google/cloud/bigtable/row_response.py
daniel-sanche: Yeah, sounds good. I changed `keys()` to `get_column_components`, and removed the others

google/cloud/bigtable/row_response.py
```diff
@@ -36,47 +37,270 @@ class RowResponse(Sequence["CellResponse"]):
...
 
-    def __init__(self, key: row_key, cells: list[CellResponse]):
+    def __init__(
+        self,
+        key: row_key,
+        cells: list[CellResponse]
+        | dict[tuple[family_id, qualifier], list[dict[str, Any]]],
+    ):
+        """
+        Initializes a RowResponse object
```
igorbernstein2, line 47: I think we should drop the Response suffix

google/cloud/bigtable/row_response.py
```diff
@@ -36,47 +37,270 @@ class RowResponse(Sequence["CellResponse"]):
     cells = row["family", "qualifier"]
     """
 
-    def __init__(self, key: row_key, cells: list[CellResponse]):
+    def __init__(
+        self,
+        key: row_key,
+        cells: list[CellResponse]
+        | dict[tuple[family_id, qualifier], list[dict[str, Any]]],
```
igorbernstein2, line 44: I think there should only be one way to represent data in a response

igorbernstein2: maybe make this internal as well

google/cloud/bigtable/row_response.py
```diff
@@ -36,47 +37,270 @@ class RowResponse(Sequence["CellResponse"]):
...
+        Initializes a RowResponse object
+
+        RowResponse objects are not intended to be created by users.
+        They are returned by the Bigtable backend.
+        """
         self.row_key = key
-        self.cells: OrderedDict[
-            family_id, OrderedDict[qualifier, list[CellResponse]]
+        self._cells_map: dict[
+            family_id, dict[qualifier, list[CellResponse]]
```
igorbernstein2, line 54: I think a better data structure would be: OrderedDict[family, list[Cell]) and use bsearch to find a qualifier in the cell list

google/cloud/bigtable/row_response.py
igorbernstein2: as discussed, endusers should never construct these

google/cloud/bigtable/row_response.py
```diff
@@ -114,17 +342,91 @@ def __int__(self) -> int:
...
-        raise NotImplementedError
+        """
+        Implements `<` operator
+        """
+        if not isinstance(other, CellResponse):
+            return NotImplemented
+        this_ordering = (
+            self.family,
+            self.column_qualifier,
+            -self.timestamp_micros,
```
igorbernstein2, line 384: please dont fight native ordering

daniel-sanche: My understanding was that newer cells should come first? By default, older timestamps would come first.

That said, comparison is less relevant if we just use cells as they come from the backend though. If we don't have a way to determine proper ordering from the client side, maybe we should remove these comparison implementations entirely?

PR #766, opened 2023-04-17
feat: sharded read rows

daniel-sanche:
Blocked on Read Rows PR: https://github.com/googleapis/python-bigtable/pull/762

This PR adds query sharding. 

- Adds `query.shard()`, to split one large query into multiple smaller ones
- Adds `client.sample_keys`, to get a list of "sample points" for a table, that are used to efficiently shard
- Adds `client.read_rows_sharded`, to execute a set of sharded queries in parallel, and return the results as a single list
- made some usability changes to RowRange and Query classes


google/cloud/bigtable/client.py
```diff
@@ -567,7 +604,20 @@ async def sample_keys(
                 will be chained with a RetryExceptionGroup containing all GoogleAPIError
                 exceptions from any retries that failed
         """
-        raise NotImplementedError
+        operation_timeout = operation_timeout or self.default_operation_timeout
```
daniel-sanche, line 605: - Do we want to raise an error if any of the shard queries overlap? Or is it ok to get duplicate rows?
- Do we want this method to retry?

google/cloud/bigtable/client.py
```diff
@@ -510,20 +520,49 @@ async def read_rows_sharded(
...
+        results_lists = await asyncio.gather(*routine_list, return_exceptions=True)
+        exception_list = [
+            FailedQueryShardError(idx, query_list[idx], e)
+            for idx, e in enumerate(results_lists)
+            if isinstance(e, Exception)
+        ]
+        if exception_list:
+            # if any sub-request failed, raise an exception instead of returning results
+            raise ShardedReadRowsExceptionGroup(exception_list, len(query_list))
+        combined_list = list(chain.from_iterable(results_lists))
```
daniel-sanche, line 564: I currently implemented it where rows from all queries are concatenated into a single list. But maybe we should return a list of lists, to keep results from each query seoarate?

google/cloud/bigtable/client.py
```diff
@@ -510,20 +520,49 @@ async def read_rows_sharded(
...
+        ```
 
         Args:
             - query_list: a list of queries to run in parallel
-        """
-        raise NotImplementedError
+        Raises:
+            - ShardedReadRowsExceptionGroup: if any of the queries failed
+            - ValueError: if the query_list is empty
+        """
```
daniel-sanche, line 544: Do we want to raise an error if any of the shard queries overlap? Or is it ok to get duplicate rows?

google/cloud/bigtable/client.py
igorbernstein2: I like the current approach, I think callers will be more interested in the full results than the splits. So I'd rather optimize for that usecase.

google/cloud/bigtable/client.py
```diff
@@ -542,32 +581,41 @@ async def row_exists(
...
         raise NotImplementedError
 
-    async def sample_keys(
+    async def sample_row_keys(
         self,
         *,
-        operation_timeout: int | float | None = 60,
-        per_sample_timeout: int | float | None = 10,
-        per_request_timeout: int | float | None = None,
+        operation_timeout: int | float | None = None,
```
igorbernstein2, line 587: Why no attempt timeout?

google/cloud/bigtable/client.py
igorbernstein2: yep

google/cloud/bigtable/client.py
```diff
@@ -542,32 +581,41 @@ async def row_exists(
...
-                exceptions from any retries that failed
+            - GoogleAPICallError: if the sample_row_keys request fails
         """
-        raise NotImplementedError
+        operation_timeout = operation_timeout or self.default_operation_timeout
+
+        if operation_timeout <= 0:
+            raise ValueError("operation_timeout must be greater than 0")
+
+        # TODO: add metadata
```
igorbernstein2, line 610: This cant be left as a todo - we need metadata for routing on the serverside

igorbernstein2 (changes requested):

google/cloud/bigtable/client.py
daniel-sanche: Yeah assuming they use the query.shard() function, that should be the case. But this method allows passing in a generic list of queries, so users may pass in overlapping queries, right?

igorbernstein2: You are right that its possible. I think we should avoid this situation, but not by throwing an error. I think we should make it impossible to happen. Perhaps we can do the following:

Create a Batch fetching context that end users create. The context will automatically call SampleRowKeys and cache the result. And maybe refresh it every X minutes.

The end user then interact with this object by passing it lists of keys and ranges that the context shards

google/cloud/bigtable/read_rows_query.py
```diff
@@ -197,18 +240,84 @@ def add_range(
...
+            # start either bisects left or right, depending on whether the range start is inclusive
+            if this_range.start is None:
+                start_segment = 0
+            elif this_range.start.is_inclusive:
+                start_segment = bisect.bisect_left(split_points, this_range.start.key)
+            else:
+                start_segment = bisect.bisect_right(split_points, this_range.start.key)
+            # split points represent end of range; always bisect left
+            end_segment = (
+                bisect.bisect_left(split_points, this_range.end.key)
```
igorbernstein2, line 291: minor optimization: set the `lo` to the start segment

google/cloud/bigtable/read_rows_query.py
```diff
@@ -197,18 +240,84 @@ def add_range(
...
+                if this_range.end is not None
+                else len(split_points)
+            )
```
igorbernstein2, line 294: This is very hard to read. It looks like bisect_left results are ignored. I think using statement form of if would make it a lot easier to read:

```
if this_range.end is not None:
   end_segment = bisect.bisect_left(split_points, this_range.end.key)
else:
   end_segment = len(split_points)
```

Also there is a small opportunity for an optimization of starting the binary search from the start_segment

google/cloud/bigtable/read_rows_query.py
```diff
@@ -197,18 +240,84 @@ def add_range(
...
+
+            # create new ranges for each segment
+            if start_segment == end_segment:
+                # range is contained in a single segment
+                sharded_queries[start_segment].add_range(this_range)
+            else:
+                # range spans multiple segments
+                # create start and end ranges
+                start_range = RowRange._from_points(
+                    this_range.start, _RangePoint(split_points[start_segment], True)
```
igorbernstein2, line 304: please use named arg for True, its hard to infer what Tru signifies here

google/cloud/bigtable/read_rows_query.py
```diff
@@ -197,18 +240,84 @@ def add_range(
...
+                start_segment = 0
+            elif this_range.start.is_inclusive:
+                start_segment = bisect.bisect_left(split_points, this_range.start.key)
+            else:
+                start_segment = bisect.bisect_right(split_points, this_range.start.key)
+            # split points represent end of range; always bisect left
+            end_segment = (
+                bisect.bisect_left(split_points, this_range.end.key)
+                if this_range.end is not None
+                else len(split_points)
```
igorbernstein2, line 293: the indices are bit hard to follow here. The start index is used as is, but the end index is +1?

google/cloud/bigtable/client.py
daniel-sanche: And then move the `read_rows_sharded(unsharded_query)` function onto the context object? Or something else? I'd be a bit hesitant to add more background tasks if we can avoid it, but we can probably work something out.

Another option that would be very simple to add would be to make `query.shard` return a custom `ShardedQuery` object that just wraps the query list, and then only accept that as input for `read_rows_sharded`. Or even simpler, just make define it as a type alias

Is this something we can create an issue for and address after the first alpha, or do you want it resolved before merging this?

google/cloud/bigtable/client.py
mutianf: go/cbt-client-debugging-steps#client-retry-settings needs to be updated :P

google/cloud/bigtable/read_rows_query.py
```diff
@@ -197,18 +241,113 @@ def add_range(
...
+            if this_range.end is None:
+                # if range is open on the right, include final segment
+                end_segment = len(split_points)
+            else:
+                # use binary search to find the segment the end key belongs to.
+                # optimization: remove keys up to start_segment from searched list,
+                # then add start_segment back to get the correct index
+                cropped_pts = split_points[start_segment:]
+                end_segment = (
+                    bisect_left(cropped_pts, this_range.end.key) + start_segment
```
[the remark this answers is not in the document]
daniel-sanche: Not for the end segment, since the `split_points` mark the inclusive ends of each segment. Whether the range includes or excludes the split point, the results will be the same (left side)

Do you think I should add a comment to call this out?

google/cloud/bigtable/client.py
```diff
@@ -534,20 +546,59 @@ async def read_rows_sharded(
...
-        """
-        raise NotImplementedError
+        Raises:
+            - ShardedReadRowsExceptionGroup: if any of the queries failed
+            - ValueError: if the query_list is empty
+        """
+        if not query_list:
+            raise ValueError("query_list must contain at least one query")
+        routine_list = [
+            self.read_rows(
```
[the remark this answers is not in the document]
igorbernstein2: I would imagine at the very least that calling read_rows would've started the attempt timer, which would negatively affect how much time the later batches have.

However on the flip side, I think it does the correct thing for operation timeout. I think we would want to defer calling read_rows until we are ready for it but adjust the operation timeout appropriately for each batch

google/cloud/bigtable/client.py
```diff
@@ -534,20 +546,59 @@ async def read_rows_sharded(
...
+            results_list.extend(batch_result)
+        # collect exceptions
+        exception_list = [
+            FailedQueryShardError(idx, query_list[idx], e)
+            for idx, e in enumerate(results_list)
+            if isinstance(e, Exception)
+        ]
+        if exception_list:
+            # if any sub-request failed, raise an exception instead of returning results
+            raise ShardedReadRowsExceptionGroup(exception_list, len(query_list))
```
igorbernstein2, line 599: Since you are taking the effort to let all of the shards finish despite the error, you might as well add the partial results in the exception

google/cloud/bigtable/read_rows_query.py
```diff
@@ -197,18 +241,135 @@ def add_range(
...
+            start_segment = bisect(split_points, orig_range.start.key)
+
+        # 2. find the index of the segment the end key belongs to
+        if orig_range.end is None:
+            # if range is open on the right, include final segment
+            end_segment = len(split_points)
+        else:
+            # use binary search to find the segment the end key belongs to.
+            # optimization: remove keys up to start_segment from searched list,
+            # then add start_segment back to get the correct index
```
igorbernstein2, line 330: why not just pass `lo=start_segment` to bisect_left?

google/cloud/bigtable/client.py
daniel-sanche: Ok, I made a custom type for `ShardedQueries`, which should discourage people from passing their own custom queries. We can discuss more advanced changes later. Let me know what you think

google/cloud/bigtable/client.py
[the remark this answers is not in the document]
daniel-sanche: Yes, `read_rows` is an async function, meaning it's a function that returns a `Coroutine` object when called. Nothing in the Coroutine is executed until it is awaited, this just gives us a reference to it. All of coroutines in a batch are started at the same time in `asyncio.gather`

Good point that the operation_timeout should be reduced between batches. I changed the code to only lock in the coroutine values at the start of each batch, and reduce the timeouts using the timeout generator helper

google/cloud/bigtable/client.py
daniel-sanche: Good point, I added a `successful_rows` field to the exception

google/cloud/bigtable/read_rows_query.py
daniel-sanche: good catch

igorbernstein2 (approved):

PR #620, opened 2022-08-04
docs: Update the docstring for ExactValueFilter

Mariatta:
It derives from ValueRegexFilter, so it still expects a regular expression.

The name itself was a misnomer, it was mostly to handle int values rather than exact values.

For filtering on matching str/byte values, we can use the ValueRangeFilter.

Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigtable/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)

Fixes googleapis/google-cloud-python#15327 🦕

PR #767, opened 2023-04-21
feat: literal value filter

daniel-sanche:
In response to https://github.com/googleapis/python-bigtable/pull/620/files#r939248378

This PR adds a new LiteralValueFilter to the v3 branch

It attempts to escape RE2 special characters in input text, rather than passing them as-is


google/cloud/bigtable/row_filters.py
```diff
@@ -494,7 +494,43 @@ class ExactValueFilter(ValueRegexFilter):
     def __init__(self, value: bytes | str | int):
         if isinstance(value, int):
             value = _PACK_I64(value)
-        super(ExactValueFilter, self).__init__(value)
+        elif isinstance(value, str):
+            value = value.encode("ascii")
```
daniel-sanche, line 498: Note: the filter will raise an exception if the input string can not be  encoded to ascii. That was how I understood the docstring text. Let me know if that's how it should be

tests/unit/test_row_filters.py
```diff
@@ -823,83 +823,83 @@ def test_value_regex_filter___repr__():
...
-    from google.cloud.bigtable.row_filters import ExactValueFilter
+    from google.cloud.bigtable.row_filters import LiteralValueFilter
     from google.cloud.bigtable_v2.types import data as data_v2_pb2
 
     value = 1
-    regex = struct.Struct(">q").pack(value)
-    row_filter = ExactValueFilter(value)
-    expected_dict = {"value_regex_filter": regex}
+    # encode numbers with double escapes for the backslashes
+    expected_byte_string = b"\\\x00\\\x00\\\x00\\\x00\\\x00\\\x00\\\x00\\\x01"
```
daniel-sanche, line 889: Is this the proper encoding for ints?

google/cloud/bigtable/row_filters.py
```diff
@@ -481,7 +481,7 @@ def to_dict(self) -> dict[str, bytes]:
         return {"value_regex_filter": self.regex}
 
 
-class ExactValueFilter(ValueRegexFilter):
+class LiteralValueFilter(ValueRegexFilter):
```
Mariatta, line 484: Do we need to go through deprecation process, to let users know that `ExactValueFilter` will be removed in v3, and that they should be using `LiteralValueFilter` instead?

daniel-sanche: It will not be removed right away, since there is a copy of `ExactValueFilter` and all other classes from the existing library in `google.cloud.bigtable.deprecated` in the `v3` branch

The `v3` deprecation process will be a bit tricky, but we will make sure that all existing classes are accessible for at least a decent transition period

igorbernstein2 (changes requested):

tests/unit/test_row_filters.py
daniel-sanche: I found and fixed [a bug in my null-byte detection](https://github.com/googleapis/python-bigtable/blob/804cc143792823977767604580169847cbfd5aff/google/cloud/bigtable/row_filters.py#L529), which should line up with your suggestion 1. That said, the [comments in the source snippet](https://github.com/google/re2/blob/70f66454c255080a54a8da806c52d1f618707f8a/re2/re2.cc#L479) say that it's unneeded with RE2, and my system tests seem to work with or without the special null case in place, for what it's worth.

Was that null byte the only issue you're pointing out, or are there other problems? I'm having trouble finding good test cases/documentation around RE2 parsing. [The snippet this was adapted from](https://github.com/google/re2/blob/70f66454c255080a54a8da806c52d1f618707f8a/re2/re2.cc#L456) seems to imply that all non-alphanum bytes can be escaped with a `\`, but I'm not seeing that backed up anywhere [in the docs](https://github.com/google/re2/wiki/Syntax).

igorbernstein2: ok, I'll trust you on this one :)

igorbernstein2 (approved):

PR #769, opened 2023-04-28
feat: implement mutate rows

daniel-sanche:
Implements and tests mutate_row and bulk_mutate_rows rpc calls, along with associated models and helpers

These functions will both return nothing on success, but will raise an exception with details about the failed mutation or group of mutations on error, after retries are attempted.

Errors are exposed as an ExceptionGroup in Python3.11+, allowing users to easily digest a group of errors that are raised together, in this case for multiple mutations, with multiple retries per mutation. We use a backwards compatible implementation to expose similar functionality in Python3.10 and lower

Note that mutation batching will come in a separate PR, as it builds off of the functions added here

---
todo:
- move test logic from test_client to test__mutate_rows?


google/cloud/bigtable/client.py
```diff
@@ -568,14 +599,67 @@ async def mutate_row(
...
+        if isinstance(mutations, Mutation):
+            mutations = [mutations]
+        request["mutations"] = [mutation._to_dict() for mutation in mutations]
+
+        if all(mutation.is_idempotent() for mutation in mutations):
+            # mutations are all idempotent and safe to retry
+            predicate = retries.if_exception_type(
+                core_exceptions.DeadlineExceeded,
+                core_exceptions.ServiceUnavailable,
+                core_exceptions.Aborted,
```
mutianf, line 627: same as above, we don't retry aborted mutation

google/cloud/bigtable/mutations.py
```diff
@@ -13,42 +13,121 @@
...
-    timestamp_ms: int | None = None
+    family: str
+    qualifier: bytes
+    new_value: bytes
+    timestamp_micros: int | None = None
+
+    def _to_dict(self) -> dict[str, Any]:
+        """Convert the mutation to a dictionary representation"""
+        # if timestamp not given, use -1 for server-side timestamp
+        timestamp = self.timestamp_micros if self.timestamp_micros is not None else -1
```
mutianf, line 49: I don't think we should implicitly set it to use server timestamps if this field is not set. If it's unset we can use the current timestamp. (java implementation: https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Mutation.java#LL155C1-L156C1). Implicitly using server side timestamp could lead to duplicate mutations.

google/cloud/bigtable/mutations.py
```diff
@@ -13,42 +13,121 @@
...
+                "family_name": self.family,
+                "column_qualifier": self.qualifier,
+                "timestamp_micros": timestamp,
+                "value": self.new_value,
+            }
+        }
+
+    def is_idempotent(self) -> bool:
+        """Check if the mutation is idempotent"""
+        return self.timestamp_micros is not None and self.timestamp_micros >= 0
```
mutianf, line 61: maybe update to self.timestamp_micros != -1 because -1 is the special case?

google/cloud/bigtable/mutations.py
```diff
@@ -13,42 +13,121 @@
...
 class DeleteAllFromRow(Mutation):
-    pass
+    def _to_dict(self) -> dict[str, Any]:
+        return {
+            "delete_from_row": {},
+        }
 
 
-@dataclass
 class BulkMutationsEntry:
```
mutianf, line 116: Can we rename this to RowMutationEntry? because all the mutations are targeting a single row https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/RowMutationEntry.java#L26-L32

mutianf (changes requested):

google/cloud/bigtable/mutations.py
daniel-sanche: Sounds good, I made the change to match Java

google/cloud/bigtable/mutations.py
daniel-sanche: My only concern is that it gets a bit confusing with `mutate_row` and `bulk_mutate_rows`. I'm only using this class for the `bulk_mutate_rows`, which is why I added the Bulk

But I'm fine with the change, especially if that's how Java does it

google/cloud/bigtable/_mutate_rows.py
```diff
@@ -0,0 +1,239 @@
...
+    table: "Table",
+    timeout_generator: Iterator[float],
+    active_dict: dict[int, "RowMutationEntry"],
+    error_dict: dict[int, list[Exception]],
+    predicate: Callable[[Exception], bool],
+    on_terminal_state: Callable[
+        [int, "RowMutationEntry", list[Exception] | None],
+        Coroutine[None, None, None],
+    ]
+    | None = None,
```
igorbernstein2: It seems like you are abusing bound args as state variables. I think the intent of this function are much better expressed as a class with instance variables to represent state.

```
class _RetryAttempt:
   def __init__(self, gapic, mutations, rpc_timeout, operation_timeout):
      self.gapic = gapic
      self.mutations = mutations
      self.remaining_indices = range(0, len(mutations))
      self.errors = {}
      self.rpc_timeout = rpc_timeout
      self.operation_timeout = operation_timeout

   def run(self):
       request = { ... }
       for i in self.remaining_indices:
          request["mutations"].append(self.mutations[i])

      new_remaining_indices = []
      try:
          results = gapic.mutate_rows(request)
          for r in results:
              orig_index = self.remaining_indices[r.index]
              entry_error = from_grpc_status(r.error)
            
              if entry_error is None:
                  pass
              else:
                  self.errors[orig_index].append(entry_error)
                  if is_idempotent(self.mutations[orig_index]) and  is_retriable(entry_error):
                      new_remaining_indices.append(orig_index)
        catch e:
            # handle full rpc error

        self.remaining_indices = new_remaining_indices
        if self.remaining_indices:
            raise IncompleteError
```

Then the operation function can instantiate it, bind its run method to the instance and decorate it with AsyncRetry

google/cloud/bigtable/_mutate_rows.py
```diff
@@ -0,0 +1,215 @@
...
+
+    Args:
+      - gapic_client: the client to use for the mutate_rows call
+      - request: A request dict containing table name, app profile id, and other details to inclide in the request
+      - mutation_entries: a list of RowMutationEntry objects to send to the server
+      - operation_timeout: the timeout to use for the entire operation, in seconds.
+      - per_request_timeout: the timeout to use for each mutate_rows attempt, in seconds.
+          If not specified, the request will run until operation_timeout is reached.
+      - on_terminal_state: If given, this function will be called as soon as a mutation entry
+            reaches a terminal state (success or failure).
```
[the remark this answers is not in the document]
daniel-sanche: Sounds good, we can re-visit in the batching PR if needed

google/cloud/bigtable/_mutate_rows.py
```diff
@@ -0,0 +1,239 @@
...
+class _MutateRowsIncomplete(RuntimeError):
+    """
+    Exception raised when a mutate_rows call has unfinished work.
+    """
+
+    pass
+
+
+async def _mutate_rows_operation(
+    gapic_client: "BigtableAsyncClient",
```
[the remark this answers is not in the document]
daniel-sanche: Using strings can help simplify imports, and avoid circular import errors (the type-checker will still import at check time)

google/cloud/bigtable/_mutate_rows.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def on_error_fn(exc):
+        if predicate(exc) and not isinstance(exc, _MutateRowsIncomplete):
+            # add this exception to list for each active mutation
+            for idx in error_dict.keys():
+                error_dict[idx].append(exc)
+            # remove non-idempotent mutations from mutations_dict, so they are not retried
+            for idx, mut in list(mutations_dict.items()):
+                if not mut.is_idempotent():
+                    mutations_dict.pop(idx)
```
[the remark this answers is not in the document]
daniel-sanche: I simplified this as part of the class refactor

google/cloud/bigtable/_mutate_rows.py
```diff
@@ -0,0 +1,239 @@
...
+        all_errors = []
+        for idx, exc_list in error_dict.items():
+            if len(exc_list) == 0:
+                raise core_exceptions.ClientError(
+                    f"Mutation {idx} failed with no associated errors"
+                )
+            elif len(exc_list) == 1:
+                cause_exc = exc_list[0]
+            else:
+                cause_exc = bt_exceptions.RetryExceptionGroup(exc_list)
```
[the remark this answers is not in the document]
daniel-sanche: I was following a similar pattern in other rpcs (including mutate_row), where it will return the exception directly if it's not retried, but wrap a group of exceptions if there are multiple failed tries. This logic is an extension of that, but adding an extra grouping layer to represent multiple entries. 

So do you think we should change just this one and have a bit of an inconsistency there, or change all rpcs to always wrap in a RetryExceptionGroup, or leave as-is?

(This may be worth addressing as an issue in the repo, since it may require changes across a couple places)

google/cloud/bigtable/_mutate_rows.py
```diff
@@ -0,0 +1,239 @@
...
+      - per_request_timeout: the timeout to use for each mutate_rows attempt, in seconds.
+          If not specified, the request will run until operation_timeout is reached.
+      - on_terminal_state: If given, this function will be called as soon as a mutation entry
+            reaches a terminal state (success or failure).
+    """
+    mutations_dict: dict[int, RowMutationEntry] = {
+        idx: mut for idx, mut in enumerate(mutation_entries)
+    }
+
+    error_dict: dict[int, list[Exception]] = {idx: [] for idx in mutations_dict.keys()}
```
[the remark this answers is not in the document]
daniel-sanche: Defaultdict could have also worked, but this will be out of date after the class refactor

google/cloud/bigtable/_mutate_rows.py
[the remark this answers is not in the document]
igorbernstein2: this seems inconsistent with the other PRs?

google/cloud/bigtable/client.py
```diff
@@ -617,19 +625,75 @@ async def mutate_row(
...
-        mutation_entries: list[BulkMutationsEntry],
+        mutation_entries: list[RowMutationEntry],
         *,
-        operation_timeout: int | float | None = 60,
-        per_request_timeout: int | float | None = None,
+        operation_timeout: float | None = 60,
+        per_request_timeout: float | None = None,
+        on_success: Callable[
+            [int, RowMutationEntry], None | Coroutine[None, None, None]
+        ]
```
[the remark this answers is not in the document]
igorbernstein2: The current server impl unary :(
And retries should fairly rare, so for the time being lets keep the surface narrow and expand when needed

google/cloud/bigtable/_mutate_rows.py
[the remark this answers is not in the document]
igorbernstein2: Lets keep as is for now and add a github issue to take another pass on exception structure

google/cloud/bigtable/_mutate_rows.py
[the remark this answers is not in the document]
igorbernstein2: We can leave as is for now, but lets take another pass later on

google/cloud/bigtable/_mutate_rows.py
```diff
@@ -0,0 +1,220 @@
...
+    retry_wrapped = retry(attempt_context.run_attempt)
+    # convert RetryErrors from retry wrapper into DeadlineExceeded errors
+    deadline_wrapped = _convert_retry_deadline(retry_wrapped, operation_timeout)
+    try:
+        # trigger mutate_rows
+        await deadline_wrapped()
+    except Exception as exc:
+        # exceptions raised by retryable are added to the list of exceptions for all unfinalized mutations
+        for idx in attempt_context.remaining_indices:
+            attempt_context.append_error(idx, exc)
```
igorbernstein2, line 105: I think it might be better to move this logic into the attempt_context wrapped in a method called finalize_state() or something similar...it will limit the logic leak

google/cloud/bigtable/_mutate_rows.py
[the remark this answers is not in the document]
igorbernstein2: We can leave as is for now, but lets take another pass later on

google/cloud/bigtable/_mutate_rows.py
```diff
@@ -0,0 +1,220 @@
...
+                # bubble up exception to be handled by retry wrapper
+                raise
+        finally:
+            self.remaining_indices = new_remaining_indices
+        # check if attempt succeeded, or needs to be retried
+        if self.remaining_indices:
+            # unfinished work; raise exception to trigger retry
+            raise _MutateRowsIncomplete
+
+    def append_error(
```
igorbernstein2, line 201: should this be prefixed with _ since its only called internally by _MutateRowsAttemptContext?

igorbernstein2: Also please rename this to something thats a bit more broadly scoped....I was very surprised to see this messing with retry indices. Maybe _handle_entry_error?

Also I think it might be nicer to capture a local copy of self.remaining_indices before sending the RPC, set a new empty list on self for remaining_indices and have this method modify that as opposed to having this method take an output arg

google/cloud/bigtable/mutations.py
```diff
@@ -13,41 +13,200 @@
...
+    @abstractmethod
+    def _to_dict(self) -> dict[str, Any]:
+        raise NotImplementedError
+
+    def is_idempotent(self) -> bool:
+        """
+        Check if the mutation is idempotent
+        If false, the mutation will not be retried
+        """
+        return True
```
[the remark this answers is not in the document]
igorbernstein2: err, nvm, its done in the subclass

google/cloud/bigtable/mutations.py
```diff
@@ -13,41 +13,200 @@
...
 class DeleteRangeFromColumn(Mutation):
     family: str
-    column_qualifier: bytes
-    start_timestamp_ms: int
-    end_timestamp_ms: int
+    qualifier: bytes
+    # None represents 0
+    start_timestamp_micros: int | None = None
+    # None represents infinity
+    end_timestamp_micros: int | None = None
```
igorbernstein2, line 141: I think timestamps should be longs?

google/cloud/bigtable/mutations.py
```diff
@@ -13,41 +13,200 @@
...
-    family: str
-    column_qualifier: bytes
-    new_value: bytes | str | int
-    timestamp_ms: int | None = None
+    def __init__(
+        self,
+        family: str,
+        qualifier: bytes | str,
+        new_value: bytes | str | int,
+        timestamp_micros: int | None = None,
```
igorbernstein2, line 82: long

google/cloud/bigtable/mutations.py
```diff
@@ -13,41 +13,200 @@
...
+class RowMutationEntry:
+    def __init__(self, row_key: bytes | str, mutations: Mutation | list[Mutation]):
+        if isinstance(row_key, str):
+            row_key = row_key.encode("utf-8")
+        if isinstance(mutations, Mutation):
+            mutations = [mutations]
+        self.row_key = row_key
+        self.mutations = tuple(mutations)
+
+    def _to_dict(self) -> dict[str, Any]:
```
igorbernstein2, line 195: maybe in a future PR it might be good to make the mutations list mutable so that end users can build up the mutation array

google/cloud/bigtable/_mutate_rows.py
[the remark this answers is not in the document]
daniel-sanche: Sounds good, I opened one here: https://github.com/googleapis/python-bigtable/issues/794

google/cloud/bigtable/mutations.py
daniel-sanche: I opened an issue to track this here: https://github.com/googleapis/python-bigtable/issues/795

google/cloud/bigtable/mutations.py
daniel-sanche: Python doesn't have a native long type. Native ints are more abstract, and have arbitrary precision: https://peps.python.org/pep-0237/

google/cloud/bigtable/_mutate_rows.py
daniel-sanche: I ended up doing some refactoring to move the attempt method and the operation start function into the same class, to mirror what we're doing on the read_rows side. I think that makes things a bit cleaner, but let me know what you think

google/cloud/bigtable/mutations.py
igorbernstein2: I'm still stuck on python 2 :(

google/cloud/bigtable/_mutate_rows.py
```diff
@@ -0,0 +1,207 @@
...
+            gapic_client.mutate_rows,
+            table_name=table.table_name,
+            app_profile_id=table.app_profile_id,
+            metadata=metadata,
+        )
+        # create predicate for determining which errors are retryable
+        self.is_retryable = retries.if_exception_type(
+            core_exceptions.DeadlineExceeded,
+            core_exceptions.ServiceUnavailable,
+            _MutateRowsIncomplete,
```
igorbernstein2, line 83: nit. add a comment explaining the source of the errors
```
            # RPC level errors
            core_exceptions.DeadlineExceeded,
            core_exceptions.ServiceUnavailable,
            # Entry level errors
            _MutateRowsIncomplete,
```

google/cloud/bigtable/_mutate_rows.py
```diff
@@ -0,0 +1,207 @@
...
+        # make gapic request
+        try:
+            result_generator = await self._gapic_fn(
+                timeout=next(self.timeout_generator),
+                entries=request_entries,
+            )
+            async for result_list in result_generator:
+                for result in result_list.entries:
+                    # convert sub-request index to global index
+                    orig_idx = active_request_indices.pop(result.index)
```
igorbernstein2, line 168: nit. pop is doing extra work here. Why not just a lookup?

google/cloud/bigtable/_mutate_rows.py
```diff
@@ -0,0 +1,207 @@
...
+                        result.status.message,
+                        details=result.status.details,
+                    )
+                    if result.status.code == 0:
+                        continue
+                    else:
+                        self._handle_entry_error(orig_idx, entry_error)
+        except Exception as exc:
+            # add this exception to list for each mutation that wasn't
+            # already handled
```
igorbernstein2, line 180: and update remaining_indices

google/cloud/bigtable/_mutate_rows.py
```diff
@@ -0,0 +1,207 @@
...
+                    orig_idx = active_request_indices.pop(result.index)
+                    entry_error = core_exceptions.from_grpc_status(
+                        result.status.code,
+                        result.status.message,
+                        details=result.status.details,
+                    )
+                    if result.status.code == 0:
+                        continue
+                    else:
+                        self._handle_entry_error(orig_idx, entry_error)
```
igorbernstein2, line 177: Please add a comment that this will update remaining_indices

igorbernstein2 (approved):
lgtm after comments are addressed

PR #770, opened 2023-05-05
feat: mutate rows batching

daniel-sanche:
Blocked on https://github.com/googleapis/python-bigtable/pull/769. This PR builds off of changes from this branch

Implements `MutationsBatcher` to submit mutations in the background.


google/cloud/bigtable/mutations_batcher.py
```diff
@@ -15,16 +15,117 @@
...
+    """
+
+    def __init__(self, max_mutation_count: int | None, max_mutation_bytes: int | None):
+        """
+        Args:
+          - max_mutation_count: maximum number of mutations to send in a single rpc.
+             This corresponds to individual mutations in a single RowMutationEntry.
+             If None, no limit is enforced.
+          - max_mutation_bytes: maximum number of bytes to send in a single rpc.
+             If None, no limit is enforced.
```
igorbernstein2, line 51: should the args be flow_control_max_count, flow_control_max_bytes?

google/cloud/bigtable/mutations_batcher.py
```diff
@@ -46,61 +147,214 @@ class MutationsBatcher:
...
+        if self.closed:
+            raise RuntimeError("Cannot append to closed MutationsBatcher")
+        size = mutations.size()
+        if size > self._flow_control.max_mutation_bytes:
+            raise ValueError(
+                f"Mutation size {size} exceeds flow_control_max_bytes: {self._flow_control.max_mutation_bytes}"
+            )
+        if len(mutations.mutations) > self._flow_control.max_mutation_count:
+            raise ValueError(
+                f"Mutation count {len(mutations.mutations)} exceeds flow_control_max_count: {self._flow_control.max_mutation_count}"
```
igorbernstein2, line 221: I think we  want flow control to be advisory rather then a hard limit. Its meant to limit the throughput of an individual client rather than prevent traffic. Having said that, I think there are separate limits that need to be enforced (100k mutations per entry and 100k entries.

So batch flush limits should be loose and force a flush even after an overflow. But the hard limits should be enforced at the model level

daniel-sanche: Ok, I changed it so that the flow control will allow oversized entries as a single-entry request, and I added some extra checks for very large entries and requests

Is there a hard limit to the bytes size of an individual entry/request I should be checking for as well as well?

google/cloud/bigtable/mutations_batcher.py
daniel-sanche: I thought that could be left out since this is the flow control class, and they will be accessed as "flow_control.max_mutation_count". But I can change it to be more explicit if you'd prefer

google/cloud/bigtable/mutations_batcher.py
```diff
@@ -46,61 +174,221 @@ class MutationsBatcher:
...
-    async def append_conditional(
+        Args:
+          - raise_exceptions: if True, will raise any unreported exceptions from this or previous flushes.
+              If False, exceptions will be stored in self.exceptions and raised on a future flush
+              or when the batcher is closed.
+          - timeout: maximum time to wait for flush to complete, in seconds.
+              If exceeded, flush will continue in the background and exceptions
+              will be surfaced on the next flush
+        Raises:
+          - MutationsExceptionGroup if raise_exceptions is True and any mutations fail
```
igorbernstein2, line 276: I dont we want anyone to ever disable exceptions here

google/cloud/bigtable/mutations_batcher.py
```diff
@@ -46,61 +174,221 @@ class MutationsBatcher:
...
+        """
+        if interval is None:
+            return
+        while not self.closed:
+            await asyncio.sleep(interval)
+            # add new flush task to list
+            if not self.closed and self._staged_mutations:
+                self._schedule_flush()
+
+    def append(self, mutations: RowMutationEntry | list[RowMutationEntry]):
```
igorbernstein2, line 235: Lets avoid adding multiple entries at a time. To support this usage we would want to break up the list into batches, which will complicate things.

google/cloud/bigtable/mutations_batcher.py
```diff
@@ -46,61 +174,221 @@ class MutationsBatcher:
...
-        Add a new mutation to the internal queue
+        if self.closed:
+            raise RuntimeError("Cannot append to closed MutationsBatcher")
+        if not isinstance(mutations, list):
+            mutations = [mutations]
+        for m in mutations:
+            if isinstance(m, Mutation):  # type: ignore
+                raise ValueError(
+                    f"invalid mutation type: {type(m).__name__}. Only RowMutationEntry objects are supported by batcher"
+                )
```
igorbernstein2, line 253: Shouldnt the type hints be enough?

google/cloud/bigtable/mutations_batcher.py
```diff
@@ -46,61 +174,221 @@ class MutationsBatcher:
...
+          - new_entries: list of mutations to flush
+          - prev_flush: the previous flush task, which will be awaited before
+              a new flush is initiated
         """
-        raise NotImplementedError
+        # wait for previous flush to complete
+        await prev_flush
+        # flush new entries
+        async for batch in self._flow_control.add_to_flow(new_entries):
+            batch_errors = await self._execute_mutate_rows(batch)
```
igorbernstein2, line 317: Doesnt this prevent concurrency? The idea of a batcher is that it should break up a stream of mutations into X concurrent batches of size Y and when X * Y limit is reached, it applies back pressure to the caller or queues the mutations.

igorbernstein2 (commented):
I added a few nit-picking comments. But I think the primary missing feature here is that there is no concurrency:
- there can only be a single flush task
- and a flush task will wait on the results of a mutate_rows result

daniel-sanche (commented):
addressed PR comments, and then added a few optimizations (reducing extra flush tasks when one is pending, raising exceptions with the proper flush call)

google/cloud/bigtable/mutations.py
```diff
@@ -17,10 +17,14 @@
 import time
 from dataclasses import dataclass
 from abc import ABC, abstractmethod
+from sys import getsizeof
 
 # special value for SetCell mutation timestamps. If set, server will assign a timestamp
 SERVER_SIDE_TIMESTAMP = -1
 
+# mutation entries above this should be rejected
+MAX_MUTATIONS_PER_ENTRY = 100_000
```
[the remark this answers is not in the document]
mutianf: Yes, it is a hard limit on the bigtable server side https://cloud.google.com/bigtable/docs/writes#simple. The limit of a single row is 256MB https://cloud.google.com/bigtable/quotas#:~:text=A%20single%20value%20in%20a%20table%20cell%3A%20100%20MB,a%20single%20row%3A%20256%20MB so I think it's probably good to limit bytes to 256 if user sets it to none.

google/cloud/bigtable/mutations_batcher.py
daniel-sanche: To be clear, any remaining exceptions will always be raise when the batcher is closed. This flag just allows users to create silent flush tasks that save their exceptions to be raised later, if they want to flush in the background somewhere.

I'm fine with removing it if that's what you want, but I do think there is a risk that some exceptions could be lost if background tasks aren't finalized properly

google/cloud/bigtable/mutations_batcher.py
```diff
@@ -46,61 +172,246 @@ class MutationsBatcher:
...
+        if self.closed:
+            raise RuntimeError("Cannot append to closed MutationsBatcher")
+        if isinstance(mutation_entry, Mutation):  # type: ignore
+            raise ValueError(
+                f"invalid mutation type: {type(mutation_entry).__name__}. Only RowMutationEntry objects are supported by batcher"
+            )
+        if self._scheduled_flush_entries:
+            # flush is already scheduled to run on next loop iteration
+            # add new entries directly to flush list
+            self._scheduled_flush_entries.append(mutation_entry)
```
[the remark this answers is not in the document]
daniel-sanche: Although I guess another way to solve the problem would be to make `append` into an async coroutine, and yield immediately to the event loop in there when the flush limits are hit. There are pros and cons, but it would make the code a bit simpler. Let me know what you think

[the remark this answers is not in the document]
daniel-sanche: I ended up making the `async append` change, since it should simplify things for the sync surface.

google/cloud/bigtable/client.py
```diff
@@ -576,17 +575,42 @@ async def sample_keys(
...
         raise NotImplementedError
 
-    def mutations_batcher(self, **kwargs) -> MutationsBatcher:
+    def mutations_batcher(
+        self,
+        *,
+        flush_interval: float | None = 5,
+        flush_limit_mutation_count: int | None = 1000,
+        flush_limit_bytes: int = 20 * MB_SIZE,
+        flow_control_max_count: int | None = 100_000,
```
igorbernstein2, line 584: should this flow_control_mutation_count?

google/cloud/bigtable/client.py
```diff
@@ -583,10 +581,22 @@ def mutations_batcher(self, **kwargs) -> MutationsBatcher:
         Can be used to iteratively add mutations that are flushed as a group,
         to avoid excess network calls
 
+        Kwargs:
+          - flush_interval: Automatically flush every flush_interval seconds. If None,
+              a table default will be used
+          - flush_limit_count: Flush immediately after flush_limit_count mutations are added.
+              If None, this limit is ignored.
+          - flush_limit_bytes: Flush immediately after flush_limit_bytes bytes are added.
+              If None, this limit is ignored.
```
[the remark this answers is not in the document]
igorbernstein2: I think we can leave it unenforced for now. And come back to it

google/cloud/bigtable/mutations_batcher.py
```diff
@@ -15,16 +15,140 @@
...
+
+    Flow limits are not hard limits. If a single mutation exceeds the configured
+    limits, it will be allowed as a single batch when the capacity is available.
+    """
+
+    def __init__(
+        self,
+        max_mutation_count: int | None,
+        max_mutation_bytes: int | None,
+    ):
```
igorbernstein2, line 53: I thought we were dropping None?

google/cloud/bigtable/client.py
daniel-sanche: Good point, fixed

google/cloud/bigtable/mutations_batcher.py
daniel-sanche: I think I was still unsure whether we wanted to fully remove the behaviour of `max... = None` (by adding a hard cap), or just discourage it by removing None from the signature.

I dropped it from the signature for now, but let me know if you think we need to add a cap too

daniel-sanche:
@igorbernstein2 @mutianf 

I [added timeout configuration](https://github.com/googleapis/python-bigtable/pull/770/commits/eaed9ab82caee34ba3f4b92a7b9cf17c07662d8d) to the batcher init signature. Previously, it would always use the table defaults. This is part of the public API, so just wanted to make sure you saw the change in case you have thoughts

google/cloud/bigtable/mutations_batcher.py
```diff
@@ -46,61 +177,318 @@ class MutationsBatcher:
...
+            while not self.closed:
+                await asyncio.sleep(interval)
+                # add new flush task to list
+                if not self.closed and self._staged_entries:
+                    self._schedule_flush()
+
+        timer_task = asyncio.create_task(timer_routine(self, interval))
+        return timer_task
+
+    async def append(self, mutation_entry: RowMutationEntry):
```
igorbernstein2, line 278: Please add a TODO to return a future with the result of this mutation

google/cloud/bigtable/mutations_batcher.py
```diff
@@ -46,61 +177,318 @@ class MutationsBatcher:
...
+          - new_entries: list of RowMutationEntry objects to flush
+        """
+        # flush new entries
+        in_process_requests: list[asyncio.Future[list[FailedMutationEntryError]]] = []
+        async for batch in self._flow_control.add_to_flow(new_entries):
+            batch_task = self._create_bg_task(self._execute_mutate_rows, batch)
+            in_process_requests.append(batch_task)
+        # wait for all inflight requests to complete
+        found_exceptions = await self._wait_for_batch_results(*in_process_requests)
+        # allow previous flush tasks to finalize before adding new exceptions to list
```
igorbernstein2, line 331: is this comment stale?

google/cloud/bigtable/mutations_batcher.py
```diff
@@ -46,61 +177,318 @@ class MutationsBatcher:
...
+        )
+        self._flush_timer = self._start_flush_timer(flush_interval)
+        self._flush_jobs: set[asyncio.Future[None]] = set()
+        # MutationExceptionGroup reports number of successful entries along with failures
+        self._entries_processed_since_last_raise: int = 0
+        self._exceptions_since_last_raise: int = 0
+        # keep track of the first and last _exception_list_limit exceptions
+        self._exception_list_limit: int = 10
+        self._oldest_exceptions: list[Exception] = []
+        self._newest_exceptions: list[Exception] = []
```
igorbernstein2, line 244: should this be a deque?

google/cloud/bigtable/mutations_batcher.py
```diff
@@ -46,61 +177,318 @@ class MutationsBatcher:
...
+        self._flush_timer.cancel()
+        self._schedule_flush()
+        await asyncio.gather(*self._flush_jobs, return_exceptions=True)
+        try:
+            await self._flush_timer
+        except asyncio.CancelledError:
+            pass
+        # raise unreported exceptions
+        self._raise_exceptions()
+        atexit.unregister(self._on_exit)
```
igorbernstein2, line 437: I think you might want to do this before raising exceptions

google/cloud/bigtable/mutations_batcher.py
```diff
@@ -46,61 +177,318 @@ class MutationsBatcher:
...
 
     async def close(self):
         """
         Flush queue and clean up resources
         """
-        raise NotImplementedError
+        self.closed = True
+        self._flush_timer.cancel()
+        self._schedule_flush()
+        await asyncio.gather(*self._flush_jobs, return_exceptions=True)
```
igorbernstein2, line 430: is it ok for gather to get an empty list?

google/cloud/bigtable/mutations_batcher.py
daniel-sanche: I don't think it should make too much of a difference for this use-case, but I changed to deque in case that changes

google/cloud/bigtable/mutations_batcher.py
daniel-sanche: good catch

google/cloud/bigtable/mutations_batcher.py
daniel-sanche: it should be, but I added an extra check

google/cloud/bigtable/mutations_batcher.py
```diff
@@ -46,61 +177,322 @@ class MutationsBatcher:
...
+        if isinstance(mutation_entry, Mutation):  # type: ignore
+            raise ValueError(
+                f"invalid mutation type: {type(mutation_entry).__name__}. Only RowMutationEntry objects are supported by batcher"
+            )
+        self._staged_entries.append(mutation_entry)
+        # start a new flush task if limits exceeded
+        self._staged_count += len(mutation_entry.mutations)
+        self._staged_bytes += mutation_entry.size()
+        if (
+            self._staged_count >= self._flush_limit_count
```
mutianf, line 301: I think we should check if staged_count + mutation_metry.mutations > self.flush_limit_count first before adding it to the staged_entries. Otherwise we could exceed the flush_limit_count?

igorbernstein2 (approved):

PR #775, opened 2023-05-11
fix: Revert "fix: Revert "Feat: Threaded MutationsBatcher""

mutianf:
Reverts googleapis/python-bigtable#773

Beam is gonna pin on 2.17 so we can revert the revert :)

### DIFF
--- a/docs/batcher.rst
+++ b/docs/batcher.rst
@@ -1,0 +1,6 @@
+Mutations Batching
+~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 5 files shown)


igorbernstein2 (approved):

Abacn:
Thanks @mutianf can we still try to figure out why monitoring infos becomes empty in google-cloud-bigtable=2.18.0 before next client library release? That is put a blocker to the next release (persumably 2.19.0)

PR #780, opened 2023-05-23
feat: read_modify_write and check_and_mutate_row

daniel-sanche:
This PR implements `read_modify_write` and `check_and_mutate_row` rpcs, along with related classes


mutianf:
Is it possible to extend the read_modify_write and check_and_mutate row to enable retry on certain error codes in the future?

mutianf (approved):
LGTM, the retries will be added in the future https://github.com/googleapis/python-bigtable/issues/787

PR #796, opened 2023-06-08
fix: add a callback function on flush_rows

mutianf:

### DIFF
--- a/google/cloud/bigtable/batcher.py
+++ b/google/cloud/bigtable/batcher.py
@@ -192,6 +192,10 @@ class MutationsBatcher(object):
     :type flush_interval: float
     :param flush_interval: (Optional) The interval (in seconds) between asynchronous flush.
         Default is 1 second.
+
+    :type flush_completed_callback: Callable = None
+    :param flush_completed_callback: (Optional) A callable funtion for handling responses 
+        after the request is flushed.
     """
 
     def __init__(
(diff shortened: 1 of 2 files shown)


google/cloud/bigtable/batcher.py
```diff
@@ -192,6 +192,10 @@ class MutationsBatcher(object):
     :type flush_interval: float
     :param flush_interval: (Optional) The interval (in seconds) between asynchronous flush.
         Default is 1 second.
+
+    :type flush_completed_callback: Callable = None
+    :param flush_completed_callback: (Optional) A callable funtion for handling responses
```
igorbernstein2: I dont think flush is correct here...from a user perspective it means all buffered elements have been sent, but that might not be the case. I think it should be batch_completed_callback\

google/cloud/bigtable/batcher.py
```diff
@@ -215,6 +220,7 @@ def __init__(
         )
         self.futures_mapping = {}
         self.exceptions = queue.Queue()
+        self.flush_completed_callback = flush_completed_callback
```
igorbernstein2, line 223: please prefix with _ to mark it as provite

google/cloud/bigtable/batcher.py
```diff
@@ -357,6 +363,9 @@ def _flush_rows(self, rows_to_flush):
         if len(rows_to_flush) > 0:
             response = self.table.mutate_rows(rows_to_flush)
 
+            if self.flush_completed_callback:
+                self.flush(response)
```
igorbernstein2, line 367: this look very wrong :)

igorbernstein2 (changes requested):

google/cloud/bigtable/batcher.py
mutianf: oops 😂

igorbernstein2 (approved):

igorbernstein2 (approved):

daniel-sanche (approved):

Issue #809, opened 2023-06-22
v3: warm channels with metadata


daniel-sanche:
PingAndWarm is currently called with only the instance id. We need to call it with proper metadata and app_profile_id to ensure warming works as intended

instance, table_name, and app_profile_id are all needed to form a proper warm request

Issue #804, opened 2023-06-21
v3: x-goog-request-params metadata should use ampersand


daniel-sanche:
currently, `x-goog-request-params` metadata is sent with a "," delimeter. Should replace with "&"

PR #810, opened 2023-06-22
feat: ping and warm with metadata

daniel-sanche:
Changes:
- Previously, ping_and_warm was only warming an instance name. Now, rpcs are sent with proper app_profile_id and metadata fields for proper routing
    - to enable this, registered instances are stored in the client as an `(instance_name, table_name, app_profile_id)` namedtuple instead of just `instance_name`. Any time a table is created with a change in any of those fields, a new instance will be registered for warming on the client
- changed _ping_and_warm_instance to allow warming channels for a single new instance, so all channels can be warmed for a new instance when added, without re-warming all existing instances
- also, fixed metadata to use `&` as delimeter instead of `,`

Fixes https://github.com/googleapis/python-bigtable/issues/809, https://github.com/googleapis/python-bigtable/issues/804

### DIFF
--- a/google/cloud/bigtable/client.py
+++ b/google/cloud/bigtable/client.py
@@ -217,7 +217,12 @@ class BigtableDataClient(ClientWithProject):
         tasks = [
             ping_rpc(
                 request={"name": instance_name, "app_profile_id": app_profile_id},
-                metadata=_make_metadata(table_name, app_profile_id),
+                metadata=[
+                    (
+                        "x-goog-request-params",
+                        f"name={instance_name}&app_profile_id={app_profile_id}",
+                    )
+                ],
                 wait_for_ready=True,
             )
             for (instance_name, table_name, app_profile_id) in instance_list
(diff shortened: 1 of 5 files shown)


google/cloud/bigtable/client.py
```diff
@@ -186,26 +193,39 @@ async def close(self, timeout: float = 2.0):
...
             "/google.bigtable.v2.Bigtable/PingAndWarm",
             request_serializer=PingAndWarmRequest.serialize,
         )
-        tasks = [ping_rpc({"name": n}) for n in self._active_instances]
-        result = await asyncio.gather(*tasks, return_exceptions=True)
+        # prepare list of coroutines to run
+        tasks = [
+            ping_rpc(
+                request={"name": instance_name, "app_profile_id": app_profile_id},
+                metadata=_make_metadata(table_name, app_profile_id),
```
igorbernstein2, line 220: I think I might've misled you earlier. metadata must always match the request's resource & app_profile.
For ReadRows, the resource key is table and the value is 
So the metadata is x-goog-params=table=projects/my-project/instances/my-instance/tables/my-table&app_profile_id=my-app-profile

For PingAndWarm the resource key is name and its value is the instance name So the metadata is x-goog-params=name=projects/my-project/instances/my-instance&app_profile_id=my-app-profile

daniel-sanche: fixed

igorbernstein2 (approved):

Issue #811, opened 2023-06-23
move existing client back to current namespace


daniel-sanche:
At the start of `v3` development, I moved the existing client into `google.cloud.bigtable.deprecated`. For the alpha, we'll need it to be back in it's original namespace.

The new client should be moved to `google.cloud.bigtable.data` to avoid name collisions with duplicated model classes

Along with this change, we should make sure required public classes are suffixed with `Async`

Issue #805, opened 2023-06-21
v3: use upstream `api_core`


daniel-sanche:
We currently use a submodule witha  custom version of `api_core` for the v3 library. We have published a snapshot as [2.12.0.dev0](https://pypi.org/project/google-api-core/2.12.0.dev0/). We should change our settings to use this version by default, to allow easier installation

Tests should still run against submodule for the time being though, for development

daniel-sanche:
This change can likely be grouped in with https://github.com/googleapis/python-bigtable/issues/803

PR #816, opened 2023-06-26
chore: restructure module paths

daniel-sanche:
This PR moves the existing client back into place in `google.cloud.bigtable`, allowing the addition of the new client to be a non-breaking change. The new client is now accessed at `google.cloud.bigtable.data`, to avoid name collisions between similar model classes. 

(Fixes https://github.com/googleapis/python-bigtable/issues/811)

---

As part of this move, async classes have been moved into `google.cloud.bigtable.data._async` subdirectory:
  - `client.py`
  - `_mutate_rows.py`
  - `mutations_batcher.py`
  - `_read_rows.py`

 This resulted in a couple classes being repositioned to keep async classes fully separate from sync helper classes:
  - `_StateMachine` moved from `_read_rows` to `_read_rows_statemachine`
  - `_MutateRowsIncomplete` error moved from `_mutate_rows` to `exceptions`
  - `ReadRowsIterator` moved from `iterators` to `_read_rows`.

 Publically-initializable async classes `BigtableDataClientAsync` and `TableAsync` are also exposed in `google.cloud.bigtable.data.__init__.py`, so end-users will not need to import anything from `google.cloud.bigtable.data._async` directly

---

This PR also pins the dependency for `python-api-core` to `2.12.0.dev0`, which has support for retryable streams. Previously, users would need to install the `python-api-core` submodule manually to use the library, but now it should be self-contained.

`2.12.0.dev0` is a [pre-release developer build on pypi](https://pypi.org/project/google-api-core/2.12.0.dev0/), so we will need to update this to the official `2.12.0` when it is ready

(Fixes https://github.com/googleapis/python-bigtable/issues/805)

---

Also merged the latest `main` back into the `v3` branch, so updates to the `main` branch will be present in the files changed view


daniel-sanche:
It looks like the failing docs tests are something upstream affecting all python libraries, not related to this change

igorbernstein2 (approved):

Issue #782, opened 2023-05-23
v3: rethink timeouts


daniel-sanche:
From comments in the read_rows PR:
- https://github.com/googleapis/python-bigtable/pull/762/files#r1198023391
```
nit: maybe rename to attempt_timeout?
```
- https://github.com/googleapis/python-bigtable/pull/762/files#r1201264529
```
 think the params for scans should be distinct from simple unary timeouts. I think maybe having:

- operation_timeout/attempt_timeout control mutate_row, check_and_mutate_row, read_modify_write, read_row
- scan_operation_timeout/scan_attempt_timeout for read_rowS
- batch_operation_timeout/attempt_operation_timeout for mutate_rowS
```

- https://github.com/googleapis/python-bigtable/pull/762/files#r1198104210
```
Can end users disable timeouts?
```

Timeouts must be propagated to mutation_batcher in an intelligent way as well

PR #819, opened 2023-06-27
feat: improve timeout structure

daniel-sanche:
- renamed `per_request_timeout` to `attempt_timeout`
- previously, we just had `default_operation_timeout` and `default_per_request_timeout` values configurable at the table level. Now, have different defaults for read_rows and mutate_rows
- centralized timeout validation into a new helper function
- improved some docstrings related to timeouts

Fixes https://github.com/googleapis/python-bigtable/issues/782


google/cloud/bigtable/data/_async/_mutate_rows.py
```diff
@@ -52,15 +52,15 @@ def __init__(
...
+        attempt_timeout: float | None,
     ):
         """
         Args:
           - gapic_client: the client to use for the mutate_rows call
           - table: the table associated with the request
           - mutation_entries: a list of RowMutationEntry objects to send to the server
           - operation_timeout: the timeout t o use for the entire operation, in seconds.
-          - per_request_timeout: the timeoutto use for each mutate_rows attempt, in seconds.
+          - attempt_timeout: the timeoutto use for each mutate_rows attempt, in seconds.
```
mutianf, line 63:
```suggestion
          - attempt_timeout: the timeout to use for each mutate_rows attempt, in seconds.
```

google/cloud/bigtable/data/_async/_mutate_rows.py
```diff
@@ -52,15 +52,15 @@ def __init__(
...
         operation_timeout: float,
-        per_request_timeout: float | None,
+        attempt_timeout: float | None,
     ):
         """
         Args:
           - gapic_client: the client to use for the mutate_rows call
           - table: the table associated with the request
           - mutation_entries: a list of RowMutationEntry objects to send to the server
           - operation_timeout: the timeout t o use for the entire operation, in seconds.
```
mutianf, line 62:
```suggestion
          - operation_timeout: the timeout to use for the entire operation, in seconds.
```

google/cloud/bigtable/data/_async/client.py
```diff
@@ -393,8 +398,12 @@ def __init__(
         table_id: str,
         app_profile_id: str | None = None,
         *,
-        default_operation_timeout: float = 600,
-        default_per_request_timeout: float | None = None,
+        default_read_rows_operation_timeout: float = 600,
+        default_read_rows_attempt_timeout: float | None = None,
+        default_mutate_rows_operation_timeout: float = 600,
+        default_mutate_rows_attempt_timeout: float | None = None,
```
mutianf, line 404: We should have default for attempt timeout. mutate_rows = 1 minute, other operations = 20 seconds.

google/cloud/bigtable/data/_async/client.py
```diff
@@ -439,7 +451,13 @@ def __init__(
         self.app_profile_id = app_profile_id
 
         self.default_operation_timeout = default_operation_timeout
-        self.default_per_request_timeout = default_per_request_timeout
+        self.default_attempt_timeout = default_attempt_timeout
+        self.default_read_rows_operation_timeout = default_read_rows_operation_timeout
+        self.default_read_rows_attempt_timeout = default_read_rows_attempt_timeout
+        self.default_mutate_rows_operation_timeout = (
+            default_mutate_rows_operation_timeout
```
mutianf, line 458: why the parentheses?

google/cloud/bigtable/data/_async/client.py
```diff
@@ -524,20 +538,37 @@ async def read_rows(
...
+                If None, defaults to the Table's default_read_rows_attempt_timeout,
+                or the operation_timeout if that is also None.
         Returns:
-            - a list of the rows returned by the query
+            - a list of Rows returned by the query
+        Raises:
+            - DeadlineExceeded: raised after operation timeout
+                will be chained with a RetryExceptionGroup containing GoogleAPIError exceptions
+                from any retries that failed
+            - GoogleAPIError: raised if the request encounters an unrecoverable error
```
mutianf, line 566: Nit: I think we should rename this to GoogleApiError. I couldn't find the style guide on this, but this seems to be the convention for some other classes, example: https://github.com/googleapis/google-cloud-python/blob/main/packages/google-cloud-api-gateway/google/cloud/apigateway_v1/types/apigateway.py#L53

google/cloud/bigtable/data/_async/mutations_batcher.py
```diff
@@ -203,25 +204,18 @@ def __init__(
           - flow_control_max_bytes: Maximum number of inflight bytes.
           - batch_operation_timeout: timeout for each mutate_rows operation, in seconds. If None,
               table default_operation_timeout will be used
-          - batch_per_request_timeout: timeout for each individual request, in seconds. If None,
-              table default_per_request_timeout will be used
+          - batch_attempt_timeout: timeout for each individual request, in seconds. If None,
+              table default_attempt_timeout will be used
```
mutianf, line 208: if None, table default mutate_rows_attempt_timeout will be used. If that's also none, table's default operation_timeout will be used.

google/cloud/bigtable/data/_async/mutations_batcher.py
```diff
@@ -203,25 +204,18 @@ def __init__(
...
-            batch_operation_timeout or table.default_operation_timeout
+            batch_operation_timeout or table.default_mutate_rows_operation_timeout
         )
-        self._per_request_timeout: float = (
-            batch_per_request_timeout
-            or table.default_per_request_timeout
+        self._attempt_timeout: float = (
+            batch_attempt_timeout
+            or table.default_mutate_rows_attempt_timeout
             or self._operation_timeout
```
mutianf, line 216: Should this be tables default_mutate_rows_operation_timeout? Because that's used as the default operation_timeout

google/cloud/bigtable/data/_async/mutations_batcher.py
```diff
@@ -203,25 +204,18 @@ def __init__(
           - flow_control_max_bytes: Maximum number of inflight bytes.
           - batch_operation_timeout: timeout for each mutate_rows operation, in seconds. If None,
               table default_operation_timeout will be used
```
mutianf, line 206: default_mutate
```suggestion
              table default_mutate_rows_operation_timeout will be used
```

google/cloud/bigtable/data/_async/mutations_batcher.py
daniel-sanche: good catch!

google/cloud/bigtable/data/_async/mutations_batcher.py
daniel-sanche: You mean instead of self._operation_timeout?

My intent was to set the attempt time as the full operation time if it's not set anywhere else. I used self._operation_timeout instead of the table default because the operation timeout may have been overridden by the batch_operation_timeout, and if so we should use that.

Or am I thinking about this wrong?

google/cloud/bigtable/data/_async/client.py
daniel-sanche: That does look better, but I'm referring to [this class](https://github.com/googleapis/python-api-core/blob/0acda8c96b23af2238ae0f4536e4fbb98bc7d787/google/api_core/exceptions.py#L59) in `api_core` here, which is used across out libraries. So we can't rename it unfortunately

google/cloud/bigtable/data/_async/client.py
daniel-sanche: That's what out linter does when it breaks long statements up into multiple lines. It looks like this is 1 char too long to fit on a single line :disappointed:

google/cloud/bigtable/data/_async/client.py
daniel-sanche: Sure, should read_rows also be 20 seconds then?

daniel-sanche: Also, previously we would raise an exception if the `attempt_timeout` was set greater than the `operation_timeout`, but I think that would be annoying if we have a default value pre-set, so I removed it. Let me know if you disagree with that

google/cloud/bigtable/data/_async/client.py
```diff
@@ -340,14 +341,18 @@ async def _remove_instance_registration(
...
         app_profile_id: str | None = None,
-        default_operation_timeout: float = 600,
-        default_per_request_timeout: float | None = None,
+        *,
+        default_read_rows_operation_timeout: float = 600,
+        default_read_rows_attempt_timeout: float | None = None,
+        default_mutate_rows_operation_timeout: float = 600,
+        default_mutate_rows_attempt_timeout: float | None = None,
+        default_operation_timeout: float = 60,
+        default_attempt_timeout: float | None = None,
```
igorbernstein2, line 355: does the pydoc need to be updated?

google/cloud/bigtable/data/_async/client.py
```diff
@@ -393,8 +398,12 @@ def __init__(
...
         app_profile_id: str | None = None,
         *,
-        default_operation_timeout: float = 600,
-        default_per_request_timeout: float | None = None,
+        default_read_rows_operation_timeout: float = 600,
+        default_read_rows_attempt_timeout: float | None = 20,
+        default_mutate_rows_operation_timeout: float = 600,
+        default_mutate_rows_attempt_timeout: float | None = 60,
+        default_operation_timeout: float = 60,
+        default_attempt_timeout: float | None = 20,
```
igorbernstein2, line 406: same as above, I think the pydoc is missing most of these

google/cloud/bigtable/data/_async/client.py
```diff
@@ -483,35 +501,31 @@ async def read_rows_stream(
             - GoogleAPIError: raised if the request encounters an unrecoverable error
             - IdleTimeout: if iterator was abandoned
         """
+        operation_timeout = (
+            operation_timeout or self.default_read_rows_operation_timeout
+        )
+        attempt_timeout = (
+            attempt_timeout
+            or self.default_read_rows_attempt_timeout
+            or operation_timeout
```
igorbernstein2, line 510: I dont think we want to fall back on operation_time for read_rows or mutate_rows

google/cloud/bigtable/data/_async/mutations_batcher.py
igorbernstein2: I dont think mutate_rows/read_rows should default to the unary timeouts

google/cloud/bigtable/data/_async/client.py
daniel-sanche: Fixed

google/cloud/bigtable/data/_async/client.py
daniel-sanche: Good catch, fixed

google/cloud/bigtable/data/_async/client.py
daniel-sanche: This is the final operation_timeout for the call, configured one line above, not the table's default_operation_out.

The purpose of this is to set the attempt_timeout == the operation_timeout if the attempt timeout remains unset, so we don't have to do None checks later on

google/cloud/bigtable/data/_async/mutations_batcher.py
daniel-sanche: `self._operation_timeout` isn't the unary timeout here, it is set as `self._operation_timeout = batch_operation_timeout or table.default_mutate_rows_operation_timeout`. 

So the attempt timeout is falling back to the operation_timeout detected for this call, either passed in directly or from the table default

igorbernstein2 (approved):

Issue #814, opened 2023-06-24
v3: think about exposing dicts


daniel-sanche:
We currently allow passing in dicts or objects for many APIs. Maybe we should force just objects for the first alpha, to avoid locking that in to the API?

PR #820, opened 2023-06-27
chore: reduce public api surface

daniel-sanche:
Turn public functions into private functions where possible, to reduce the scope of the public surface. 

This includes removing dict argument support for in places where we currently accept objects or dicts. Eventually, we should still aim support dict input to be aligned with other GCP libraries, but can come after the alpha

Includes some changes to the `RowRange` model class surface to make sure internal `_RangePoint` is kept private

Fixes: https://github.com/googleapis/python-bigtable/issues/814


google/cloud/bigtable/data/read_rows_query.py
```diff
@@ -48,11 +53,24 @@ def __init__(
         start_is_inclusive: bool | None = None,
         end_is_inclusive: bool | None = None,
     ):
+        """
+        Args:
+          - start_key: The start key of the range. If None, the range is unbounded on the left.
+          - end_key: The end key of the range. If None, the range is unbounded on the right.
+          - start_is_inclusive: Whether the start key is inclusive. If None, the start key is
+                inclusive.
```
igorbernstein2, line 61: Why not set the default value to True? Instead of default being None and then using an if?

google/cloud/bigtable/data/read_rows_query.py
```diff
@@ -66,29 +84,62 @@ def __init__(
             end_key = end_key.encode()
         elif end_key is not None and not isinstance(end_key, bytes):
             raise ValueError("end_key must be a string or bytes")
+        # ensure that start_key is less than or equal to end_key
+        if start_key is not None and end_key is not None and start_key > end_key:
```
igorbernstein2, line 88: It seems like there are 2 ways to represent infinite start/end. The bigtable native way is an empty string. We now also have None. It would good to normalize this in the constructor

google/cloud/bigtable/data/read_rows_query.py
```diff
@@ -123,7 +174,35 @@ def __bool__(self) -> bool:
...
+    def __str__(self) -> str:
+        """
+        Represent range as a string, e.g. "[b'a', b'z)"
+        Unbounded start or end keys are represented as "-inf" or "+inf"
+        """
+        left = "[" if self.start_is_inclusive else "("
+        right = "]" if self.end_is_inclusive else ")"
+        start = repr(self.start_key) if self.start_key is not None else "-inf"
+        end = repr(self.end_key) if self.end_key is not None else "+inf"
+        return f"{left}{start}, {end}{right}"
```
igorbernstein2, line 193: Can we normalize the open bounds? ie we shouldn't have both (-inf and [-inf

igorbernstein2: Perhaps the ctor can ensure that {start,end}_is_inclusive is deterministically set when the value is open ended

google/cloud/bigtable/data/read_rows_query.py
daniel-sanche: This way lets us know if they explicitly passed in a value. If we used a default, we won't know if they set it True manually, or if it's falling back to the default.

I think the reason we did this here was because it doesn't really make sense to set the inclusive flag when the key is None, so we raise an error. This was the behaviour of the existing [RowFilter class](https://github.com/googleapis/python-bigtable/blob/62dd21798a5c646fb5628e47d484cfb8a594de50/google/cloud/bigtable/row_filters.py#L375)

google/cloud/bigtable/data/read_rows_query.py
daniel-sanche: Ok, good catch. Do you mean convert empty strings to None internally?

google/cloud/bigtable/data/read_rows_query.py
daniel-sanche: `_is_inclusive` will be None when the range is None, so it will currently always be "(" and ")"

If we want to say that unbounded points always count as either inclusive or exclusive, that could make things simpler. But I wasn't sure if that could have downstream consequences

google/cloud/bigtable/data/read_rows_query.py
igorbernstein2: I dont know if I agree, but I dont want to block this pr. We can always change it

google/cloud/bigtable/data/read_rows_query.py
igorbernstein2: yep

google/cloud/bigtable/data/read_rows_query.py
igorbernstein2: Downstream it sholdnt matter because it will end up in the unset bound oneof case

daniel-sanche: Ok, I made empty ranges count as inclusive

google/cloud/bigtable/data/read_rows_query.py
daniel-sanche: Yeah I'm not too attached to either way. I opened a bug for future discussion: https://github.com/googleapis/python-bigtable/issues/826

igorbernstein2 (approved):

PR #842, opened 2023-07-24
chore: add user agent suffix

daniel-sanche:
This PR adds a suffix to the `gccl` user-agent string, so we can tell the difference between the legacy client and the v3 data client in our metrics.

The legacy client is remains `X.Y.Z`, but the v3 client is `X.Y.Z-data-async`. Eventually, we can also have `X.Y.Z-data-sync`

### DIFF
--- a/google/cloud/bigtable/client.py
+++ b/google/cloud/bigtable/client.py
@@ -161,7 +161,7 @@ class Client(ClientWithProject):
     ):
         if client_info is None:
             client_info = client_info_lib.ClientInfo(
-                client_library_version=f"{bigtable.__version__}-legacy",
+                client_library_version=bigtable.__version__,
             )
         if read_only and admin:
             raise ValueError(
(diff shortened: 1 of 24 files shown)


google/cloud/bigtable/client.py
```diff
@@ -161,7 +161,7 @@ def __init__(
     ):
         if client_info is None:
             client_info = client_info_lib.ClientInfo(
-                client_library_version=bigtable.__version__,
+                client_library_version=f"{bigtable.__version__}-legacy",
```
igorbernstein2, line 164: Thinking about this a bit more, Lets leave the old agent the same in case there are dashboard that care about it. Instead have the new clients be suffixed with sync & async

daniel-sanche: Sounds good, reverted

mutianf (approved):

PR #863, opened 2023-09-07
chore: optimize gapic calls

daniel-sanche:
This PR pulls in changes from https://github.com/googleapis/gapic-generator-python/pull/1755 and https://github.com/googleapis/python-api-core/pull/527 to improve rpc performance. 

I measured this change as giving a 6% throughput boost to the point read benchmark in a prod environment

Fixes https://github.com/googleapis/python-bigtable/pull/863

### DIFF
--- a/google/cloud/bigtable_v2/services/bigtable/async_client.py
+++ b/google/cloud/bigtable_v2/services/bigtable/async_client.py
@@ -285,13 +285,13 @@ class BigtableAsyncClient:
 
         # Wrap the RPC method; this adds retry and timeout information,
         # and friendly error handling.
-        rpc = self._client._transport._wrapped_methods[self._client._transport.read_rows]
+        rpc = self._client._transport._wrapped_methods[
+            self._client._transport.read_rows
+        ]
         # Certain fields should be provided within the metadata header;
         # add these here.
         metadata = tuple(metadata) + (
-            _cached_to_metadata(
-                (("table_name", request.table_name),)
-            ),
+            _cached_to_metadata((("table_name", request.table_name),)),
         )
 
         # Send the request.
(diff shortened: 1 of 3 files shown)


google/cloud/bigtable_v2/services/bigtable/async_client.py
```diff
@@ -1194,9 +1164,7 @@ def read_change_stream(
         # Certain fields should be provided within the metadata header;
         # add these here.
         metadata = tuple(metadata) + (
-            gapic_v1.routing_header.to_grpc_metadata(
-                (("table_name", request.table_name),)
-            ),
+            _cached_to_metadata((("table_name", request.table_name),)),
```
igorbernstein2, line 1167: Unless I'm mistaken in almost all cases metadata will start with being empty, why not include it in the cache key as well?

google/cloud/bigtable_v2/services/bigtable/transports/grpc_asyncio.py
```diff
@@ -512,6 +514,66 @@ def read_change_stream(
             )
         return self._stubs["read_change_stream"]
 
+    def _prep_wrapped_messages(self, client_info):
```
igorbernstein2, line 517: how is is this called?

google/cloud/bigtable_v2/services/bigtable/async_client.py
daniel-sanche: My end goal was to get this fix upstreamed in api-core, so I was trying to keep the signature the same. 

I ended up [getting that through](https://github.com/googleapis/python-api-core/pull/526) though, so I removed this part from this PR

google/cloud/bigtable_v2/services/bigtable/transports/grpc_asyncio.py
daniel-sanche: It's overriding the [base class implementation](https://github.com/googleapis/python-bigtable/blob/94bfe664c5e9daa3bc76d6f423dd12b7949b2f76/google/cloud/bigtable_v2/services/bigtable/transports/base.py#L130) to use `gapic_v1.method_async.wrap_method` instead of `gapic_v1.method.wrap_method`

igorbernstein2 (approved):

PR #870, opened 2023-10-05
chore(tests): add conformance tests to CI for v3

daniel-sanche:
adds config to run conformance tests pre-submit

### DIFF
--- a/.github/workflows/conformance.yaml
+++ b/.github/workflows/conformance.yaml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       matrix:
         py-version: [ 3.8 ]
-        client-type: [ "v3", "Legacy" ]
+        client-type: [ "Async v3", "Legacy" ]
     name: "${{ matrix.client-type }} client / python ${{ matrix.py-version }}"
     steps:
     - uses: actions/checkout@v3
--- a/.github/workflows/conformance.yaml
+++ b/.github/workflows/conformance.yaml
@@ -1,0 +1,49 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# Github action job to test core java library features on
+# downstream client libraries before they are released.
+on:
+  push:
+    branches:
+    - main
+  pull_request:
+name: conformance
+jobs:
+  conformance:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        py-version: [ 3.7, 3.11 ]
+        use-legacy-client: [ "false", "true" ]
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/checkout@v3
+      with:
+        repository: googleapis/cloud-bigtable-clients-test
+        ref: main
+        path: cloud-bigtable-clients-test
+    - uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.py-version }}
+    - uses: actions/setup-go@v4
+      with:
+        go-version: '>=1.20.2'
+    - run: chmod +x .kokoro/conformance.sh
+    - run: pip install -e .
+    - run: go version
+    - run: .kokoro/conformance.sh
+      name: "Conformance Tests"
+      env:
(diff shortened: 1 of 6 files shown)


.github/workflows/conformance.yaml
```diff
@@ -0,0 +1,54 @@
...
+    branches:
+    - main
+  pull_request:
+name: Conformance
+jobs:
+  conformance:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        py-version: [ 3.8 ]
```
danieljbruce, line 27: Should we test more versions?

.kokoro/conformance.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# Build and start the proxy in a separate process
+PROXY_PORT=9999
+pushd test_proxy
+nohup python test_proxy.py --port $PROXY_PORT $PROXY_ARGS &
+proxyPID=$!
+popd
+
+# Run the conformance test
+pushd cloud-bigtable-clients-test/tests
```
danieljbruce, line 39: Same comment as you made on my PR (https://github.com/googleapis/nodejs-bigtable/pull/1349#discussion_r1348017520). The push and pop are not necessary if you stay at the root.

.kokoro/conformance.sh
```diff
@@ -0,0 +1,48 @@
...
+eval "go test -v -proxy_addr=:$PROXY_PORT $TEST_ARGS"
+RETURN_CODE=$?
+popd
+
+# Stop the proxy
+kill $proxyPID
```
danieljbruce, line 45: Same comment Baha made here: https://github.com/googleapis/nodejs-bigtable/pull/1349#discussion_r1349784449. Maybe we could do this in a trap so that the process always exits.

danieljbruce (commented):
A few comments. Looking pretty good!

.github/workflows/conformance.yaml
daniel-sanche: Python's case is a bit complicated, since we have 2 (and soon 3) different clients we need to test. I thought it would make sense to just test the lowest Python version for now, to prevent the number of tests exploding too much. But we can easily add more in the future

.kokoro/conformance.sh
daniel-sanche: My comment was that `pushd .` is unnecessary if you aren't planning on changing directories

In this case, I am using pushd to move to the `./cloud-bigtable-clients-test/tests` folder for the next commands, before popping back to the root

.kokoro/conformance.sh
daniel-sanche: Good catch, fixed

.kokoro/conformance.sh
danieljbruce: Sounds good

google/cloud/bigtable/data/_async/client.py
```diff
@@ -132,6 +132,10 @@ def __init__(
         client_options = cast(
             Optional[client_options_lib.ClientOptions], client_options
         )
+        self._emulator_host = os.getenv(BIGTABLE_EMULATOR)
+        if self._emulator_host is not None and credentials is None:
+            # use insecure channel if emulator is set
+            credentials = google.auth.credentials.AnonymousCredentials()
```
danieljbruce, line 138: How are the credentials related to adding conformance tests?

daniel-sanche: Because this is the first time they are being run in a CI environment without GCP authentication, which was raising an error even in emulator mode

test_proxy/handlers/client_handler_legacy.py
```diff
@@ -79,54 +79,157 @@ async def ReadRows(self, request, **kwargs):
             row_list.append(dict_val)
         return row_list
 
+    @client_handler.error_safe
```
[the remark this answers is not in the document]
daniel-sanche: Yes, the conformance tests run through the test proxy. This is filling in some gaps in the test proxy implementation for the legacy client

google/cloud/bigtable/data/_async/client.py
danieljbruce: Looks good to me. It seems like this situation is a lot like something I had to do in datastore for setting insecure credentials when the emulator is being used. See https://github.com/googleapis/nodejs-datastore/pull/1164/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R516. I guess this is a bug fix as well then?

I might recommend in a future PR to also use the insecure credentials if an endpoint is provided in the constructor to hit the emulator when this environment variable is not provided.

test_proxy/handlers/client_handler_legacy.py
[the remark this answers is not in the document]
danieljbruce: Sounds good. Changes here are pretty low risk since they don't affect client behavior so I guess they can be worked on over time until all test cases are handled.

danieljbruce:
LGTM

PR #872, opened 2023-10-06
chore(tests): add conformance tests

daniel-sanche:
This PR replicates https://github.com/googleapis/python-bigtable/pull/870 in the main branch

To accomplish this, we needed to pull in the test_proxy code from `experimental_v3`

Note that the conformance tests are currently failing. We will need to address this in a future PR


test_proxy/handlers/grpc_handler.py
```diff
@@ -0,0 +1,148 @@
+
```
kolea2, line 1: missing header?

test_proxy/test_proxy.py
```diff
@@ -0,0 +1,193 @@
...
+
+    # Listen to requests from grpc server process
+    print_msg = "client_handler_process started"
+    if use_legacy_client:
+        print_msg += ", using legacy client"
+    print(print_msg)
+    client_map = {}
+    background_tasks = set()
+    while True:
+        if not request_q.empty():
```
danieljbruce, line 108: I'm still trying to understand the architecture of this PR. Why is the functionality of CreateClient, CloseClient and RemoveClient handled here instead of being delegated down to the next level like all the other functions sort of like how the node.js test proxy works? (ie. https://github.com/googleapis/nodejs-bigtable/blob/79035298bbf663c98d46c95021f7c944ef2c97f2/testproxy/services/index.js#L38-L47)

daniel-sanche: Yes, it is annoyingly complicated because of a quirk of python protobuf. We can't load gapic proto objects and raw proto objects in the same Python instance, so we need to do a complex multi-processing system that passes messages through a queue to accomplish anything here.

The original test_proxy PR is [here](https://github.com/googleapis/python-bigtable/pull/836), which has more discussuion if you're interested, and I tried to describe some of the architecture in [the docstrings](https://github.com/daniel-sanche/python-bigtable/blob/50648705b88b75219ae418979427dabdbe1b1cec/test_proxy/test_proxy.py#L22). But I suggest you ignore all the test_proxy code that was already reviewed there, this is just moving the same code into the `main` branch

test_proxy/handlers/grpc_handler.py
daniel-sanche: good catch!

daniel-sanche:
We are currently focusing on the conformance tests for the v3 version, and will come back to fix this one in the future. I'm going to close this for now.

Issue #779, opened 2023-05-22
v3: replace dicts with protos


daniel-sanche:
There are some places in the v3 API surface where we are using dicts as the canonical representation for data ([example](https://github.com/daniel-sanche/python-bigtable/blob/3921093359be4de47322e4667486742fb24e11d3/google/cloud/bigtable/_read_rows.py#L72)). Dicts make it easier for users to pass in configuration without constructing custom objects, but they make it harder to properly verify types and make assertions about the state of the request

Before release, we should do a pass through the code to see if there are any places where protos would be a better fit than dicts.

We should also consider whether unstructured dicts should be accepted in the a public API surface (as is standard for GCP libraries), or if we should lean more towards type safety for python-bigtable

mutianf:
Estimate: 2 days

daniel-sanche:
I went through the files and found dict requests in the following places:

- [ ] mutate_row
- [ ] check_and_mutate
- [ ] read_modify_write
- [ ] MutateRowsOperationAsync

PR #875, opened 2023-10-16
feat: replace internal dictionaries with protos in gapic calls

daniel-sanche:
Replace dicts with protos in the following rpc calls:
- [x] mutate_row
- [x] check_and_mutate
- [x] read_modify_write
- [x] MutateRowsOperationAsync

This will help us catch errors with easier type checking, and should have minor performance benefits

Fixes https://github.com/googleapis/python-bigtable/issues/779

### DIFF
--- a/google/cloud/bigtable/data/_async/_mutate_rows.py
+++ b/google/cloud/bigtable/data/_async/_mutate_rows.py
@@ -104,7 +104,7 @@ class _MutateRowsOperationAsync:
         self.timeout_generator = _attempt_timeout_generator(
             attempt_timeout, operation_timeout
         )
-        self.mutations = mutation_entries
+        self.mutations = [(m, m._to_pb()) for m in mutation_entries]
         self.remaining_indices = list(range(len(self.mutations)))
         self.errors: dict[int, list[Exception]] = {}
 
@@ -135,7 +135,7 @@ class _MutateRowsOperationAsync:
                     cause_exc = exc_list[0]
                 else:
                     cause_exc = bt_exceptions.RetryExceptionGroup(exc_list)
-                entry = self.mutations[idx]
+                entry = self.mutations[idx][0]
                 all_errors.append(
                     bt_exceptions.FailedMutationEntryError(idx, entry, cause_exc)
                 )
@@ -154,7 +154,7 @@ class _MutateRowsOperationAsync:
           - GoogleAPICallError: if the gapic rpc fails
         """
         request_entries = [
-            self.mutations[idx]._to_dict() for idx in self.remaining_indices
+            self.mutations[idx][1] for idx in self.remaining_indices
         ]
         # track mutations in this request that have not been finalized yet
         active_request_indices = {
@@ -213,7 +213,7 @@ class _MutateRowsOperationAsync:
           - idx: the index of the mutation that failed
           - exc: the exception to add to the list
         """
-        entry = self.mutations[idx]
+        entry = self.mutations[idx][0]
         self.errors.setdefault(idx, []).append(exc)
         if (
             entry.is_idempotent()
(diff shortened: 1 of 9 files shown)


google/cloud/bigtable/data/_async/client.py
```diff
@@ -943,24 +943,19 @@ async def mutate_row(
...
+        if not mutations:
+            raise ValueError("No mutations provided")
+        mutations_list = mutations if isinstance(mutations, list) else [mutations]
 
-        if isinstance(mutations, Mutation):
-            mutations = [mutations]
-        request["mutations"] = [mutation._to_dict() for mutation in mutations]
-
-        if all(mutation.is_idempotent() for mutation in mutations):
+        if all(mutation.is_idempotent() for mutation in mutations_list):
```
igorbernstein2, line 958: please leave a todo to change this to any so that we can retry the things that are retryable

google/cloud/bigtable/data/mutations.py
```diff
@@ -36,6 +39,12 @@ class Mutation(ABC):
     def _to_dict(self) -> dict[str, Any]:
         raise NotImplementedError
 
+    def _to_pb(self) -> data_pb.Mutation:
```
igorbernstein2, line 42: Do we need to double store all of the attributes? Can't we store all of the attributes in the proto directly?
Also can we drop _to_dict altogether?

igorbernstein2 (commented):
Can the model just wrap the proto directly? then you won't have to have _EntryWithProto to memoize the pb

google/cloud/bigtable/data/mutations.py
daniel-sanche: I think the main downside is we'd need a lot more boilerplate setters/getters (we lose the simplicity of the dataclasses). And marshaling to/from the protos is more expensive, but I guess that should'nt be too much of an issue here. I see this is how we handled it in the [ReadRowsQuery class](https://github.com/googleapis/python-bigtable/blob/experimental_v3/google/cloud/bigtable/data/read_rows_query.py#L184), which is more complicated than these would be.

What do you think of making these immutable? That would simplify the setters/getters, and then we wouldn't have to worry about making a static copy before starting a mutate_rows operation

daniel-sanche: > Also can we drop _to_dict altogether?

I think so. We do use `row._to_dict` in the test proxy, but I don't think we'd need it for mutations

google/cloud/bigtable/data/_async/_mutate_rows.py
```diff
@@ -104,7 +116,7 @@ def __init__(
         self.timeout_generator = _attempt_timeout_generator(
             attempt_timeout, operation_timeout
         )
-        self.mutations = mutation_entries
+        self.mutations = [_EntryWithProto(m, m._to_pb()) for m in mutation_entries]
```
daniel-sanche, line 119: If we make the mutation classes immutable, we can simplify this

google/cloud/bigtable/data/_async/client.py
daniel-sanche: I thought mutate_row was meant to be treated as a single atomic unit. Wouldn't retrying individual mutations mess up the ordering?

igorbernstein2: my mistake

google/cloud/bigtable/data/_async/_mutate_rows.py
igorbernstein2: I think making each individual mutation (ie SetCell, DeleteCell, etc) is a great idea
Howevr I dont think you can make the collection of mutations immutable (ie RowMutationEntry)

google/cloud/bigtable/data/mutations.py
igorbernstein2: ideally the logic for converting from models to protos for the test proxy is self contained in the proxy. A side goal of test proxy was to be a rosetta stone for expressing the same concepts across every cleint impl

igorbernstein2 (approved):
I think the pr is an improvement in its current state. Feel free to address the remaining comments in a follow up pr

daniel-sanche:
Sure, opened a new PR for this here: https://github.com/googleapis/python-bigtable/issues/887

Issue #787, opened 2023-05-24
v3: expose retryable errors to users


daniel-sanche:
Currently, retryable errors are configured entirely internally. We should let users override the defaults

should be exposed as error codes, not exceptions

- [x] read rows
- [x] mutate rows
- [x] mutate row
- [x] sample row keys
- [x] mutation batcher
- [x] set at table level

mutianf:
Estimate: 2 days

PR #879, opened 2023-10-20
feat: expose retryable error codes to users

daniel-sanche:
This PR adds the option to set retryable error codes at the rpc level, or at the table level

Users can donfigure retryable errors using grpc.StatusCode objects, int codes corresponding to a grpc status, or an Exception type

Fixes: https://github.com/googleapis/python-bigtable/issues/787

Blocked on https://github.com/googleapis/python-bigtable/pull/880; if we decide to change how table defaults are configured, that will have implications for this change as well

### DIFF
--- a/google/cloud/bigtable/data/_async/client.py
+++ b/google/cloud/bigtable/data/_async/client.py
@@ -497,6 +497,8 @@ class TableAsync:
         Raises:
           - RuntimeError if called outside of an async context (no running event loop)
         """
+        # NOTE: any changes to the signature of this method should also be reflected
+        # in client.get_table()
         # validate timeouts
         _validate_timeouts(
             default_operation_timeout, default_attempt_timeout, allow_none=True
--- a/google/cloud/bigtable/data/_async/client.py
+++ b/google/cloud/bigtable/data/_async/client.py
@@ -1047,8 +1047,8 @@ class TableAsync:
                 Can be passed as a sequence of grpc.StatusCodes, int representations, or
                 the corresponding GoogleApiCallError Exception types.
                 Only idempotent mutations will be retried.
-                If None, uses the Table's default_read_rows_retryable_error_codes, which defaults
-                to 4 (DeadlineExceeded), 14 (ServiceUnavailable), and 10 (Aborted)
+                If None, uses the Table's default_retryable_error_codes, which defaults
+                to 4 (DeadlineExceeded), and 14 (ServiceUnavailable)
         Raises:
              - DeadlineExceeded: raised after operation timeout
                  will be chained with a RetryExceptionGroup containing all
@@ -1075,7 +1075,7 @@ class TableAsync:
         if all(mutation.is_idempotent() for mutation in mutations):
             # mutations are all idempotent and safe to retry
             predicate = retries.if_exception_type(
-                *_errors_from_codes(retryable_error_codes, self.default_mutate_rows_retryable_error_codes)
+                *_errors_from_codes(retryable_error_codes, self.default_retryable_error_codes)
             )
         else:
             # mutations should not be retried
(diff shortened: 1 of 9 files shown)


google/cloud/bigtable/data/_async/client.py
```diff
@@ -366,21 +370,10 @@ async def _remove_instance_registration(
...
-        app_profile_id: str | None = None,
-        *,
-        default_read_rows_operation_timeout: float = 600,
-        default_read_rows_attempt_timeout: float | None = None,
-        default_mutate_rows_operation_timeout: float = 600,
-        default_mutate_rows_attempt_timeout: float | None = None,
-        default_operation_timeout: float = 60,
-        default_attempt_timeout: float | None = None,
-    ) -> TableAsync:
+    def get_table(self, instance_id: str, table_id: str, *args, **kwargs) -> TableAsync:
```
igorbernstein2, line 373: How do endusers know what args & kwargs they can pass here?

google/cloud/bigtable/data/_async/client.py
```diff
@@ -402,15 +395,23 @@ def get_table(
...
         """
-        return TableAsync(
-            self,
-            instance_id,
-            table_id,
-            app_profile_id,
-            default_operation_timeout=default_operation_timeout,
-            default_attempt_timeout=default_attempt_timeout,
-        )
+        return TableAsync(self, instance_id, table_id, *args, **kwargs)
```
igorbernstein2, line 414: can you add a comment to TableAsync ctor that reminds people to update this doc when adding args?

google/cloud/bigtable/data/_async/client.py
```diff
@@ -442,6 +443,16 @@ def __init__(
         default_mutate_rows_attempt_timeout: float | None = 60,
         default_operation_timeout: float = 60,
         default_attempt_timeout: float | None = 20,
+        default_read_rows_retryable_errors: Sequence[
+            grpc.StatusCode | int | type[Exception]
+        ] = (DeadlineExceeded, ServiceUnavailable, Aborted),
+        default_mutate_rows_retryable_errors: Sequence[
+            grpc.StatusCode | int | type[Exception]
+        ] = (DeadlineExceeded, ServiceUnavailable),
+        default_retryable_errors: Sequence[grpc.StatusCode | int | type[Exception]] = (
```
igorbernstein2, line 452: I would suggest to pick one way to express error types and stick with it. I dont think there much to gain from this flexibility

google/cloud/bigtable/data/_async/client.py
```diff
@@ -566,13 +609,16 @@ async def read_rows(
         *,
         operation_timeout: float | TABLE_DEFAULT = TABLE_DEFAULT.READ_ROWS,
         attempt_timeout: float | None | TABLE_DEFAULT = TABLE_DEFAULT.READ_ROWS,
+        retryable_errors: Sequence[grpc.StatusCode | int | type[Exception]]
+        | TABLE_DEFAULT = TABLE_DEFAULT.READ_ROWS,
```
igorbernstein2, line 613: Should TABLE_DEFAULT enum be prefixed with an underscore? My understanding is that its used as a marker value to signal that the user passed something in. I dont think it makes sense for endusers to be able to set it

igorbernstein2 (approved):
lgtm after comments are addressed

google/cloud/bigtable/data/_async/client.py
daniel-sanche: I can see arguments either way. TABLE_DEFAULT already exists in the codebase for handling timeouts though, so if we want to change it, maybe it should be a separate PR?

google/cloud/bigtable/data/_async/client.py
daniel-sanche: Are you ok with sticking with `type[Exception]`? That's what `api-core` uses in retries, so it would be simplest. But I thought you mentioned that it should accept StatusCodes at one point

PR #896, opened 2023-12-11
fix: mutations batcher race condition

daniel-sanche:
There was a race condition found in the threaded mutations batcher. The batcher would check the state of the queue, and then read from the queue as separate steps, which left room for other threads to change the state in between

This PR addresses the problem with the following changes:
-  `queue.get` will now return None if the queue is empty. This allows us to check the state and get the next items in an atomic operation, replacing the need for `queue.empty`
- rewrote the `_flush_async` implementation, to simplify it and make it more thread-safe

### DIFF
--- a/google/cloud/bigtable/batcher.py
+++ b/google/cloud/bigtable/batcher.py
@@ -347,7 +347,9 @@ class MutationsBatcher(object):
             self.flow_control.control_flow(batch_info)
             future = self._executor.submit(self._flush_rows, rows_to_flush)
             # schedule release of resources from flow control
-            future.add_done_callback(partial(self._batch_completed_callback, copy(batch_info)))
+            future.add_done_callback(
+                partial(self._batch_completed_callback, copy(batch_info))
+            )
 
     def _batch_completed_callback(self, batch_info, future):
         """Callback for when the mutation has finished to clean up the current batch
--- a/google/cloud/bigtable/batcher.py
+++ b/google/cloud/bigtable/batcher.py
@@ -18,8 +18,6 @@ import queue
 import concurrent.futures
 import atexit
 
-from functools import partial
-from copy import copy
 
 from google.api_core.exceptions import from_grpc_status
 from dataclasses import dataclass
@@ -225,6 +223,7 @@ class MutationsBatcher(object):
             max_mutations=MAX_OUTSTANDING_ELEMENTS,
             max_mutation_bytes=MAX_OUTSTANDING_BYTES,
         )
+        self.futures_mapping = {}
         self.exceptions = queue.Queue()
         self._user_batch_completed_callback = batch_completed_callback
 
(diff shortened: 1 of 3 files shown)


google/cloud/bigtable/batcher.py
```diff
@@ -51,36 +53,54 @@ def __init__(self, max_mutation_bytes=MAX_MUTATION_SIZE, flush_count=FLUSH_COUNT
...
+            except queue.Empty:
+                return None
+
+    def get_all(self):
+        """Get all items from the queue."""
+        with self._lock:
+            items = []
+            while not self._queue.empty():
+                items.append(self._queue.get_nowait())
+            self.total_mutation_count = 0
```
mutianf, line 80: should this be len(items)? can line 80 and 81 be moved out of the lock?

google/cloud/bigtable/batcher.py
```diff
@@ -51,36 +53,54 @@ def __init__(self, max_mutation_bytes=MAX_MUTATION_SIZE, flush_count=FLUSH_COUNT
...
-        self.total_size -= mutation_size
-        return row
+        """
+        Retrieve an item from the queue. Recalculate queue size.
+
+        If the queue is empty, return None.
+        """
+        with self._lock:
+            try:
+                row = self._queue.get_nowait()
```
mutianf, line 66: I think we only need to lock on the queue, can the rest of the logic be moved out of the lock?

google/cloud/bigtable/batcher.py
```diff
@@ -51,36 +53,54 @@ def __init__(self, max_mutation_bytes=MAX_MUTATION_SIZE, flush_count=FLUSH_COUNT
...
 
         mutation_count = len(item._get_mutations())
 
-        self._queue.put(item)
+        with self._lock:
+            self._queue.put(item)
 
-        self.total_size += item.get_mutations_size()
-        self.total_mutation_count += mutation_count
+            self.total_size += item.get_mutations_size()
```
mutianf, line 92: same here, can these 2 lines be moved out of the lock?

google/cloud/bigtable/batcher.py
```diff
@@ -303,57 +320,66 @@ def _flush_async(self):
...
-
-                # reset and start a new batch
-                rows_to_flush = []
-                mutations_size = 0
-                rows_count = 0
-                mutations_count = 0
-                batch_info = _BatchInfo()
-
-    def _batch_completed_callback(self, future):
+                rows_to_flush.append(next_row)
```
mutianf, line 337: Do we need to reset this and the other variables after flush is done?

daniel-sanche: They should ne reset at the start of the loop (line 315)

google/cloud/bigtable/batcher.py
daniel-sanche: This function was for emptying out the queue, so it reset the counts to 0. I ended up removing this function with the lock though

google/cloud/bigtable/batcher.py
daniel-sanche: I removed the lock, let me know what you think

igorbernstein2 (approved):
lgtm

PR #899, opened 2023-12-12
fix: add lock to flow control

daniel-sanche:
The MutationBatcher's _FlowControl class currently can experience a race condition, where the release of resources can be clobbered by a different thread adding resources to the flow. This can result in deadlock, where the flow is empty, but remains blocked waiting on mutations that are already complete

This PR adds a mutex lock, ensuring that only one thread has access to the flow control state at a time.


google/cloud/bigtable/batcher.py
```diff
@@ -128,8 +129,9 @@ def control_flow(self, batch_info):
         Calculate the resources used by this batch
         """
 
-        self.inflight_mutations += batch_info.mutations_count
-        self.inflight_size += batch_info.mutations_size
+        with self._lock:
+            self.inflight_mutations += batch_info.mutations_count
+            self.inflight_size += batch_info.mutations_size
         self.set_flow_control_status()
```
daniel-sanche, line 135: I believe we can keep set_flow_control_status out of the lock. Even if it occasionally gets clobbered, every `add` and `release` ends with another call to set_flow_control_status, so it should be eventually correct.

igorbernstein2 (approved):

PR #918, opened 2024-01-16
feat: support reversed scans

ron-gal:


daniel-sanche (changes requested):
There's a whole complicated state-machine in play here, so I don't think it's going to be quite this simple to get this feature working unfortunately (there are checks to make sure each row key is larger than the last one, for example)

If you're willing to take on the work of updating the state machine, the first step would probably be to add end-to-end system tests to this PR to ensure the output with `reversed=True` matches what we'd expect to see.

But if you don't have the capacity to take that on, feel free to close this and I can get to it when I have a chance

ron-gal:
> There's a whole complicated state-machine in play here, so I don't think it's going to be quite this simple to get this feature working unfortunately (there are checks to make sure each row key is larger than the last one, for example)
> 
> If you're willing to take on the work of updating the state machine, the first step would probably be to add end-to-end system tests to this PR to ensure the output with `reversed=True` matches what we'd expect to see.
> 
> But if you don't have the capacity to take that on, feel free to close this and I can get to it when I have a chance

ooh, unlucky. Okay, closing. Thanks!

PR #923, opened 2024-01-26
feat: client side metrics data model

daniel-sanche:
This PR adds the data model for the client-side metrics system

Follow-up PRs:
- https://github.com/googleapis/python-bigtable/pull/924
- https://github.com/googleapis/python-bigtable/pull/925

---

## Design

The main architecture looks like this:

bt_metrics drawio

Most of the work is done by the ActiveOperationMetric class, which is instantiated with each rpc call, and updated through the lifecycle of the call. When the rpc is complete, it will call `on_operation_complete` and `on_attempt_complete` on the MetricsHandler, which can then log the completed data into OpenTelemetry (or theoretically, other locations if needed)

Note that there are separate classes for active vs completed metrics (`ActiveOperationMetric`, `ActiveAttemptMetric`, `CompletedOperationMetric`, `CompletedAttemptMetric`). This is so that we can keep fields mutable and optional while the request is ongoing, but pass down static immutable copies once the attempt is completed and no new data is coming


google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,514 @@
...
+    utc: datetime.datetime = field(
+        default_factory=lambda: datetime.datetime.now(datetime.timezone.utc)
+    )
+    monotonic: float = field(default_factory=time.monotonic)
+
+
+class OperationType(Enum):
+    """Enum for the type of operation being performed."""
+
+    READ_ROWS = "ReadRows"
```
mutianf, line 71: I guess this is a question for Igor. In java we tag ReadRow and ReadRows differently, I wonder if we should do the same for python? Or leave it in sync with the server metrics?

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,514 @@
...
+
+
+@dataclass(frozen=True)
+class CompletedAttemptMetric:
+    """
+    An immutable dataclass representing the data associated with a
+    completed rpc attempt.
+    """
+
+    start_time: datetime.datetime
```
mutianf, line 95: Should we include OperationType here also?

mutianf, line 95: Should we include OperationType, cluster id, zone id, is streaming? Or those will be tagged on the metric later?

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,514 @@
...
+            and self.state != OperationState.CREATED
+        ):
+            return self._handle_error(
+                INVALID_STATE_ERROR.format("start_attempt", self.state)
+            )
+
+        # find backoff value
+        if self.backoff_generator and len(self.completed_attempts) > 0:
+            # find the attempt's backoff by sending attempt number to generator
+            backoff = self.backoff_generator.send(len(self.completed_attempts) - 1)
```
mutianf, line 209: would this cause overhead cause we're caculating the backoff in metrics? Are there jitters on the backoff time? Would the actual backoff time between attempts be different?

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,514 @@
...
+                    self._handle_error(
+                        f"Failed to decode {BIGTABLE_METADATA_KEY} metadata: {blob!r}"
+                    )
+        # SERVER_TIMING_METADATA_KEY should give a string with the server-latency headers
+        timing_header = cast(str, metadata.get(SERVER_TIMING_METADATA_KEY))
+        if timing_header:
+            timing_data = SERVER_TIMING_REGEX.match(timing_header)
+            if timing_data and self.active_attempt:
+                # convert from milliseconds to seconds
+                self.active_attempt.gfe_latency = float(timing_data.group(1)) / 1000
```
mutianf, line 248: Why do we convert it to seconds?

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,514 @@
...
+        Parse the response metadata blob and return a dictionary of key-value pairs.
+
+        Function is cached to avoid parsing the same blob multiple times.
+
+        Args:
+          - blob: the metadata blob as extracted from the grpc call
+        Returns:
+          - a tuple of zone and cluster_id, or None if parsing failed
+        """
+        try:
```
mutianf, line 263: This should be returned in the ResponseParams proto and you can access zoneId and clusterId from the ResponseParams proto, instead of decode the blob

google/cloud/bigtable/data/_metrics/data_model.py
daniel-sanche: I thought that's essentially what the streaming flag was for?

google/cloud/bigtable/data/_metrics/data_model.py
daniel-sanche: These are the fields that are tracked for each individual attempt. Aren't those fields all tracked at the operation level?

google/cloud/bigtable/data/_metrics/data_model.py
daniel-sanche: (Oops, looks like I left the implementation for the backoff generator [in the wrong PR](https://github.com/googleapis/python-bigtable/blob/20260dca391cc124c673f8f5ba0735dcb3b33feb/google/cloud/bigtable/data/_helpers.py#L101))

Python libraries share backoff generator code [from api_core](https://github.com/googleapis/python-api-core/blob/5922f2b8f382ee1fd9b69fafb2eed39db93b60fb/google/api_core/retry/retry_base.py#L119), which uses randomness and jitter

What I did for our custom generator is add a list to keep track of the values that were used. Then, we can retrieve values from the history. When we call `send(idx)` here, we're asking for the value it used for attempt `idx`. 

So to answer your questions, not much overhead, and we're getting the value used, not a new random one

google/cloud/bigtable/data/_metrics/data_model.py
daniel-sanche: I'm not sure if there was a reason. Are all the latency fields in millis?

google/cloud/bigtable/data/_metrics/data_model.py
daniel-sanche: I don't remember seeing this in the response. How is it sent?

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,514 @@
...
+
+import google.cloud.bigtable.data.exceptions as bt_exceptions
+
+if TYPE_CHECKING:
+    from google.cloud.bigtable.data._metrics.handlers._base import MetricsHandler
+
+
+# by default, exceptions in the metrics system are logged,
+# but enabling this flag causes them to be raised instead
+ALLOW_METRIC_EXCEPTIONS = os.getenv("BIGTABLE_METRICS_EXCEPTIONS", False)
```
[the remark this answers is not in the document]
daniel-sanche: Yeah, this is for raising an exception if anything goes wrong in the metrics system, instead of just logging. Could be useful for debugging, but I'm also fine with just removing it before merging

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,514 @@
...
+LOGGER = (
```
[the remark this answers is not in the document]
daniel-sanche: Sure, changed

google/cloud/bigtable/data/_metrics/data_model.py
daniel-sanche: closing since we discuss this in the comment below

google/cloud/bigtable/data/_metrics/data_model.py
daniel-sanche: Oh, I think I see what you mean, construct a proto using the blob?

google/cloud/bigtable/data/_metrics/data_model.py
daniel-sanche: I converted all the variables to explicitly specify ms

google/cloud/bigtable/data/_metrics/data_model.py
mutianf: Yes, these are operation level labels and should be the same across multiple attempts. But Attempts are also labeled with these fields, so want to make sure they're added to the attributes later :)

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,517 @@
...
+            return self._handle_error(
+                INVALID_STATE_ERROR.format("end_attempt_with_status", self.state)
+            )
+        if isinstance(status, Exception):
+            status = self._exc_to_status(status)
+        duration_seconds = time.monotonic() - self.active_attempt.start_time.monotonic
+        new_attempt = CompletedAttemptMetric(
+            start_time=self.active_attempt.start_time.utc,
+            first_response_latency_ms=self.active_attempt.first_response_latency_ms,
+            duration_ms=duration_seconds * 1000,
```
mutianf, line 318: is it possible to measure it in nano seconds? seconds precision seems too low :(

google/cloud/bigtable/data/_metrics/data_model.py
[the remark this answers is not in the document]
daniel-sanche: Sure, done

google/cloud/bigtable/data/_metrics/data_model.py
daniel-sanche: duration_seconds is actually higher precision than seconds already, because it's a float value.

The docs [say this about the precision](https://docs.python.org/3/library/time.html#time.monotonic): "Use [monotonic_ns()](https://docs.python.org/3/library/time.html#time.monotonic_ns) to avoid the precision loss caused by the [float](https://docs.python.org/3/library/functions.html#float) type."

So I think we should already be at sub-millisecond percision, but if that's not good enough we can change everything to `monotonic_ns` to get full int nanoseconds everywhere

google/cloud/bigtable/data/_metrics/data_model.py
daniel-sanche: Right now, `on_attempt_complete(attempt, operation)` is passed a `CompletedAttemptMetric` and an `ActiveOperationMetric`, and between the two of them, the attempt callback will have access to all the fields we need. I'm hesitant to duplicate data between them, because it means more tests and more room for errors (not to mention more slow Python lines that need to be executed)

My thinking is that some data lives at the individual attempt level, and some is shared at the operation level, so I tried to structure the data models in that way. But let me know if I'm misunderstanding something

mutianf (commented):
LGTM after nits

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,517 @@
...
+
+@dataclass(frozen=True)
+class TimeTuple:
+    """
+    Tuple that holds both the utc timestamp to record with the metrics, and a
+    monotonic timestamp for calculating durations. The monotonic timestamp is
+    preferred for calculations because it is resilient to clock changes, eg DST
+    """
+
+    utc: datetime.datetime = field(
```
mutianf, line 63: since we're only measuring latencies, why do we need the utc timestamp?

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,517 @@
...
+                else:
+                    self._handle_error(
+                        f"Failed to decode {BIGTABLE_METADATA_KEY} metadata: {blob!r}"
+                    )
+        # SERVER_TIMING_METADATA_KEY should give a string with the server-latency headers
+        timing_header = cast(str, metadata.get(SERVER_TIMING_METADATA_KEY))
+        if timing_header:
+            timing_data = SERVER_TIMING_REGEX.match(timing_header)
+            if timing_data and self.active_attempt:
+                self.active_attempt.gfe_latency_ms = float(timing_data.group(1))
```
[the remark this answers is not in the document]
mutianf: I see this is added in the other pr.

google/cloud/bigtable/data/_metrics/data_model.py
daniel-sanche: Sure, just converted everything to int nanos

mutianf (dismissed):

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,495 @@
...
+        Args:
+          - status: The status of the attempt.
+        """
+        if self.state != OperationState.ACTIVE_ATTEMPT or self.active_attempt is None:
+            return self._handle_error(
+                INVALID_STATE_ERROR.format("end_attempt_with_status", self.state)
+            )
+        if isinstance(status, Exception):
+            status = self._exc_to_status(status)
+        new_attempt = CompletedAttemptMetric(
```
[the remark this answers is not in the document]
daniel-sanche: sure, renamed to `completed_attempt`

google/cloud/bigtable/data/_helpers.py
```diff
@@ -97,6 +98,25 @@ def _attempt_timeout_generator(
...
+    """
+    Build a generator for exponential backoff sleep times.
+
+    This implementation builds on top of api_core.retries.exponential_sleep_generator,
+    adding the ability to retrieve previous values using the send(idx) method. This is
+    used by the Metrics class to track the sleep times used for each attempt.
+    """
+    history = []
+    subgenerator = exponential_sleep_generator(initial, multiplier, maximum)
+    while True:
```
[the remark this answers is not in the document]
igorbernstein2: This seems a bit too surprising of an api. I think it would be a lot cleaner and easier to read if Attempt.start() took a delay parameter

google/cloud/bigtable/data/_metrics/data_model.py
igorbernstein2: I think java ReadRow tag was my mistake :( 
Lets keep this inline with server metrics. I think the correct path forward will be stanley's request profiles

igorbernstein2 (changes requested):

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,495 @@
...
+        if (
+            self.state != OperationState.BETWEEN_ATTEMPTS
+            and self.state != OperationState.CREATED
+        ):
+            return self._handle_error(
+                INVALID_STATE_ERROR.format("start_attempt", self.state)
+            )
+
+        # find backoff value
+        if self.backoff_generator and len(self.completed_attempts) > 0:
```
igorbernstein2, line 193: nit. `self.backoff_generator and self.completed_attempts`

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,495 @@
...
+
+@dataclass
+class ActiveOperationMetric:
+    """
+    A dataclass representing the data associated with an rpc operation that is
+    currently in progress. Fields are mutable and may be optional.
+    """
+
+    op_type: OperationType
+    backoff_generator: BackoffGenerator | None = None
```
igorbernstein2, line 142: Might be good to add a comment explaining when this is None ... I'm guessing for non-retriable operations?

Separately I'm not sure that this will work for RetryInfo in response trailers (where the server specifies how much to sleep). Might be better to just pass the amount slept as an arg to start_attempt

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,495 @@
...
+    # keep monotonic timestamps for active operations
+    start_time_ns: int = field(default_factory=time.monotonic_ns)
+    active_attempt: ActiveAttemptMetric | None = None
```
igorbernstein2, line 145: When would this be non-None?

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,495 @@
...
+    cluster_id: str | None = None
+    zone: str | None = None
+    completed_attempts: list[CompletedAttemptMetric] = field(default_factory=list)
+    is_streaming: bool = False  # only True for read_rows operations
```
igorbernstein2, line 149: Thats not entirely true, it would also be true for CDC if we were to support that in this client and we have a couple of other features that would set this to true. I would remove the comment so it doesnt get stale

google/cloud/bigtable/data/_metrics/metrics_controller.py
```diff
@@ -0,0 +1,60 @@
...
+        """
+        self.handlers.append(handler)
+
+    def create_operation(
+        self, op_type: OperationType, **kwargs
+    ) -> ActiveOperationMetric:
+        """
+        Creates a new operation and registers it with the subscribed handlers.
+        """
+        handlers = self.handlers + kwargs.pop("handlers", [])
```
igorbernstein2, line 58: whats the usecase for adding a handler per operation?

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,495 @@
...
+          - status: The status of the operation.
+        """
+        if self.state == OperationState.COMPLETED:
+            return self._handle_error(
+                INVALID_STATE_ERROR.format("end_with_status", self.state)
+            )
+        final_status = (
+            self._exc_to_status(status) if isinstance(status, Exception) else status
+        )
+        if self.state == OperationState.ACTIVE_ATTEMPT:
```
igorbernstein2, line 332: under what circumstances will the state be different?

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,495 @@
...
+
+        Causes on_operation_completed to be called for each registered handler.
+        """
+        return self.end_with_status(StatusCode.OK)
+
+    def build_wrapped_predicate(
+        self, inner_predicate: Callable[[Exception], bool]
+    ) -> Callable[[Exception], bool]:
+        """
+        Wrapps a predicate to include metrics tracking. Any call to the resulting predicate
```
igorbernstein2, line 362: nit s/Wrapps/Wrap

Im having a hard time wrapping my head around this....whats an example of a predicate that will be wrapped?

igorbernstein2, line 362: nit s/Wrapps/Wrap

Im having a hard time wrapping my head around this....whats an example of a predicate that will be wrapped?

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,495 @@
...
+        Exception groups and wrappers will be parsed to find the underlying
+        grpc Exception.
+
+        If the exception is not a grpc exception, will return StatusCode.UNKNOWN.
+
+        Args:
+          - exc: The exception to extract the status code from.
+        """
+        if isinstance(exc, bt_exceptions._BigtableExceptionGroup):
+            exc = exc.exceptions[-1]
```
igorbernstein2, line 394: Why the last one? please add a note

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,495 @@
...
+        if hasattr(exc, "grpc_status_code") and exc.grpc_status_code is not None:
+            return exc.grpc_status_code
+        if (
+            exc.__cause__
+            and hasattr(exc.__cause__, "grpc_status_code")
+            and exc.__cause__.grpc_status_code is not None
+        ):
+            return exc.__cause__.grpc_status_code
```
igorbernstein2, line 402: is a single level enough? should this be recursive?

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,495 @@
...
+            """
+            Pass through trailing metadata to the wrapped operation
+            """
+            return self.operation.add_response_metadata(metadata)
+
+        def wrap_attempt_fn(
+            self,
+            fn: Callable[..., Any],
+            *,
+            extract_call_metadata: bool = True,
```
igorbernstein2, line 459: why would this be false?

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,495 @@
...
+        ) -> Callable[..., Any]:
+            """
+            Wraps a function call, tracing metadata along the way
```
igorbernstein2, line 462: Does this wrap an attempt or an operation?

google/cloud/bigtable/data/_metrics/data_model.py
igorbernstein2: Looking at the instrumentation pr, it seems like this will be used to wrap the predicate thats passed to the retry infrastructure. This feels a bit hacky...I would expect predicates to be side effect free and this relies on the implementation details of the current retry infrastructure.

I think it would be cleaner to capture retryable attempt results using the on_error_fn param in retries and to capture the overall op status using a try/catch block....it would convey the intent much more clearly

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,495 @@
...
+    An immutable dataclass representing the data associated with a
+    completed rpc attempt.
+
+    Operation-level fields (eg. type, cluster, zone) are stored on the
+    corresponding CompletedOperationMetric or ActiveOperationMetric object.
+    """
+
+    duration_ns: int
+    end_status: StatusCode
+    first_response_latency_ns: int | None = None
```
daniel-sanche, line 83: should this be moved to the Operation Metrics side?

PR #924, opened 2024-01-26
feat: client side metrics handlers

daniel-sanche:
This PR builds off of https://github.com/googleapis/python-bigtable/pull/923 to add handlers to the client-side metrics system, which can subscribe to the metrics stream, and export the results into different collection systems

Follow-up PR:
- https://github.com/googleapis/python-bigtable/pull/925

---

We add three handlers to the system:
- `GoogleCloudMetricsHandler`: sends metrics to a private OpenTelemetry meter, and then periodically exports them to GCP. Built on top of `OpenTelemetryMetricsHandler`
- `OpenTelemetryMetricsHandler`: sends metrics to the root MeterProvider, so the user can access the exported metrics for their own systems. This will be off by default, but can be added alongside `GoogleCloudMetricsHandler` if needed
- `_StdoutMetricsHandler`: can print metrics to stdout as they arrive. Mostly for debugging (we can remove this if you don't think it's useful)


google/cloud/bigtable/data/_metrics/handlers/opentelemetry.py
```diff
@@ -0,0 +1,196 @@
...
+            unit="ms",
+        )
+        self.attempt_latencies = meter.create_histogram(
+            name="attempt_latencies",
+            description="A distribution of the latency of each client RPC, tagged by operation name and the attempt status. Under normal circumstances, this will be identical to operation_latencies. However, when the client receives transient errors, operation_latencies will be the sum of all attempt_latencies and the exponential delays.",
+            unit="ms",
+        )
+        self.retry_count = meter.create_counter(
+            name="retry_count",
+            description="A count of additional RPCs sent after the initial attempt. Under normal circumstances, this will be 1.",
```
mutianf, line 59: Under normal circumstances, this value is empty. https://cloud.google.com/bigtable/docs/client-side-metrics-descriptions#retry-count

google/cloud/bigtable/data/_metrics/handlers/opentelemetry.py
```diff
@@ -0,0 +1,196 @@
...
+    def __init__(self, meter_provider=None):
+        if meter_provider is None:
+            # use global meter provider
+            from opentelemetry import metrics
+
+            meter_provider = metrics
+        # grab meter for this module
+        meter = meter_provider.get_meter("bigtable.googleapis.com")
+        # create instruments
+        self.operation_latencies = meter.create_histogram(
```
mutianf, line 42: let's make sure the metric descriptions are identical to our public doc https://cloud.google.com/bigtable/docs/client-side-metrics-descriptions

google/cloud/bigtable/data/_metrics/handlers/opentelemetry.py
```diff
@@ -0,0 +1,196 @@
...
+        client_uid: str | None = None,
+        instruments: _OpenTelemetryInstruments = _OpenTelemetryInstruments(),
+        **kwargs,
+    ):
+        super().__init__()
+        self.otel = instruments
+        # fixed labels sent with each metric update
+        self.shared_labels = {
+            "client_name": f"python-bigtable/{bigtable_version}",
+            "client_uid": client_uid or str(uuid4()),
```
mutianf, line 114: Is it possible to detect the hostname where the client is run similar to java? https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable-stats/src/main/java/com/google/cloud/bigtable/stats/BigtableStackdriverExportUtils.java#L158-L173 Or that's not possible with python?

google/cloud/bigtable/data/_metrics/handlers/opentelemetry.py
```diff
@@ -0,0 +1,196 @@
...
+            "resource_project": project_id,
+            "resource_instance": instance_id,
+            "resource_table": table_id,
+        }
+        if app_profile_id:
+            self.shared_labels["app_profile"] = app_profile_id
```
mutianf, line 120: else: we should tag app_profile with "default"

google/cloud/bigtable/data/_metrics/handlers/opentelemetry.py
```diff
@@ -0,0 +1,196 @@
...
+    def on_operation_complete(self, op: CompletedOperationMetric) -> None:
+        """
+        Update the metrics associated with a completed operation:
+          - operation_latencies
+          - retry_count
+        """
+        try:
+            status = str(op.final_status.value[0])
+        except (IndexError, TypeError):
+            status = "2"  # unknown
```
mutianf, line 131: we actually export status string instead of the numeric value, so OK, DEADLINE_EXCEEDED, etc.

google/cloud/bigtable/data/_metrics/handlers/opentelemetry.py
```diff
@@ -0,0 +1,196 @@
...
+            "resource_zone": op.zone,
+            "resource_cluster": op.cluster_id,
+            **self.shared_labels,
+        }
+        is_streaming = str(op.is_streaming)
+
+        self.otel.operation_latencies.record(
+            op.duration, {"streaming": is_streaming, **labels}
+        )
+        self.otel.retry_count.add(len(op.completed_attempts) - 1, labels)
```
mutianf, line 144: We leave retry_count as empty if there's no retries. So we only export this if (len(op.completed_attempt) - 1 > 0)

google/cloud/bigtable/data/_metrics/handlers/opentelemetry.py
```diff
@@ -0,0 +1,196 @@
...
+          - retry_count
+        """
+        try:
+            status = str(op.final_status.value[0])
+        except (IndexError, TypeError):
+            status = "2"  # unknown
+        labels = {
+            "method": op.op_type.value,
+            "status": status,
+            "resource_zone": op.zone,
```
mutianf, line 135: Do we also need to fallback to default zone and cluster here?

google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
```diff
@@ -0,0 +1,268 @@
...
+        Write a set of metrics to Cloud Monitoring.
+        This method is called by the OpenTelemetry SDK
+        """
+        deadline = time.time() + (timeout_millis / 1000)
+        metric_kind = MetricDescriptor.MetricKind.CUMULATIVE
+        all_series: list[TimeSeries] = []
+        # process each metric from OTel format into Cloud Monitoring format
+        for resource_metric in metrics_data.resource_metrics:
+            for scope_metric in resource_metric.scope_metrics:
+                for metric in scope_metric.metrics:
```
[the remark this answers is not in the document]
daniel-sanche: This exporter is attached to a private MeterProvider, so there shouldn't be any other metrics showing up here

google/cloud/bigtable/data/_metrics/handlers/opentelemetry.py
daniel-sanche: It should be possible to do something like that, I'll take a look

google/cloud/bigtable/data/_metrics/handlers/opentelemetry.py
daniel-sanche: We do that when building CompletedOperationMetric before calling `on_operation_complete`

The idea that ActiveOperationMetric has some possibly empty fields, but when it's finalized into a CompletedOperationMetric, all the defaults are applied and it becomes more type-strict

google/cloud/bigtable/data/_metrics/handlers/opentelemetry.py
daniel-sanche: Ok, so don't even send anything for this metric if there's no errors? And I assume that applies to connectivity_error_count too then?

google/cloud/bigtable/data/_metrics/handlers/opentelemetry.py
daniel-sanche: fixed, although I didn't include this part of the client blocking latencies, since it doesn't apply: `For versions 2.21.0 and later, this metric also includes the latencies of requests queued on gRPC channels.`

google/cloud/bigtable/data/_metrics/handlers/opentelemetry.py
daniel-sanche: done

google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
```diff
@@ -0,0 +1,268 @@
...
+      - connectivity_error_count: number of attempts that failed to reach Google's network.
+      - application_latencies: the time spent waiting for the application to process the next response.
+      - throttling_latencies: latency introduced by waiting when there are too many outstanding requests in a bulk operation.
+
+    Args:
+      - project_id: The Google Cloud project ID for the associated Bigtable Table
+      - export_interval: The interval (in seconds) at which to export metrics to Cloud Monitoring.
+    """
+
+    def __init__(self, *args, project_id: str, export_interval=60, **kwargs):
```
mutianf, line 105: I think the minimum interval we can publish metrics to cloud monitoring is 60 seconds. And I don't think we want to update with larger intervals. So let's remove this option?

google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
```diff
@@ -0,0 +1,268 @@
...
+
+
+# create OpenTelemetry views for Bigtable metrics
+# avoid reformatting into individual lines
+# fmt: off
+MILLIS_AGGREGATION = view.ExplicitBucketHistogramAggregation(
+    [
+        0, 0.01, 0.05, 0.1, 0.3, 0.6, 0.8, 1, 2, 3, 4, 5, 6, 8, 10, 13, 16,
+        20, 25, 30, 40, 50, 65, 80, 100, 130, 160, 200, 250, 300, 400,
+        500, 650, 800, 1000, 2000, 5000, 10000, 20000, 50000, 100000,
```
mutianf, line 58: We updated the buckets in java:

          ImmutableList.of(
              0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 8.0, 10.0, 13.0, 16.0, 20.0, 25.0, 30.0, 40.0,
              50.0, 65.0, 80.0, 100.0, 130.0, 160.0, 200.0, 250.0, 300.0, 400.0, 500.0, 650.0,
              800.0, 1000.0, 2000.0, 5000.0, 10000.0, 20000.0, 50000.0, 100000.0, 200000.0,
              400000.0, 800000.0, 1600000.0, 3200000.0));

I think 100k is too small

google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
```diff
@@ -0,0 +1,268 @@
...
+    def __init__(self, *args, project_id: str, export_interval=60, **kwargs):
+        # internal exporter to write metrics to Cloud Monitoring
+        exporter = _BigtableMetricsExporter(project_id=project_id)
+        # periodically executes exporter
+        gcp_reader = PeriodicExportingMetricReader(
+            exporter, export_interval_millis=export_interval * 1000
+        )
+        # use private meter provider to store instruments and views
+        meter_provider = MeterProvider(metric_readers=[gcp_reader], views=VIEW_LIST)
+        otel = _OpenTelemetryInstruments(meter_provider=meter_provider)
```
mutianf, line 114: how would a customer provide their own otel instrumentation? (this could be in a follow up PR)

In java we let customers override it in the settings and pass the otel instance down. see the description in https://github.com/googleapis/java-bigtable/pull/1796

google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
```diff
@@ -0,0 +1,268 @@
...
+        for resource_metric in metrics_data.resource_metrics:
+            for scope_metric in resource_metric.scope_metrics:
+                for metric in scope_metric.metrics:
+                    for data_point in [
+                        pt for pt in metric.data.data_points if pt.attributes
+                    ]:
+                        if data_point.attributes:
+                            project_id = data_point.attributes.get("resource_project")
+                            if not isinstance(project_id, str):
+                                # we expect string for project_id field
```
mutianf, line 158: Maybe log a warning? Maybe something like: "malformatted resource_project x. Skip publishing"

google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
```diff
@@ -0,0 +1,268 @@
...
+    Each GoogleCloudMetricsHandler will maintain its own exporter instance associated with the
+    project_id it is configured with.
+
+    Args:
+      - project_id: GCP project id to associate metrics with
+    """
+
+    def __init__(self, project_id: str):
+        super().__init__()
+        self.client = MetricServiceClient()
```
mutianf, line 134: do we need to configure retry attempt number for create_service_time_series method? what's the default? In java we are not retrying the method, I think because republishing could lead to errors.

google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
```diff
@@ -0,0 +1,268 @@
...
+        self.prefix = "bigtable.googleapis.com/internal/client"
+        self.project_name = self.client.common_project_path(project_id)
+
+    def export(
+        self, metrics_data: MetricsData, timeout_millis: float = 10_000, **kwargs
```
mutianf, line 139: why not just pass in second? :) so we don't need to convert it back to seconds on line 145 later :)

google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
daniel-sanche: The [client](https://github.com/googleapis/python-bigtable/blob/fcbdda667f0f888b184209aee7ab5c43f7edaee5/google/cloud/bigtable/data/_async/client.py#L532) holds a [controller](https://github.com/googleapis/python-bigtable/blob/fcbdda667f0f888b184209aee7ab5c43f7edaee5/google/cloud/bigtable/data/_metrics/metrics_controller.py#L30C1-L31C1) that manages a [set of handlers](https://github.com/googleapis/python-bigtable/tree/client_side_metrics_instrumentation/google/cloud/bigtable/data/_metrics/handlers). Users can add their own [OpenTelemetryHandler](https://github.com/googleapis/python-bigtable/blob/client_side_metrics_instrumentation/google/cloud/bigtable/data/_metrics/handlers/opentelemetry.py) to send metrics to a different MeterProvider if they want

I'll have to write up some documentation for this at some point

google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
daniel-sanche: Good point. I just looked at the gapic code, and it looks like this rpc defaults to no retries. We can add them if we want, but if we export every 60 seconds, maybe we should add back to the queue for the next batch?

Would republishing lead to errors here too?

google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
daniel-sanche: This is part of the superclass method definition unfortunately. We inherit from `opentelemetry.sdk.metrics.export.MetricExporter`

google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py
daniel-sanche, line 134: should this be created with a project?

daniel-sanche:
closing in favor of https://github.com/googleapis/python-bigtable/pull/1189

PR #925, opened 2024-01-26
feat: added client-side instrumentation to all rpcs

daniel-sanche:
This PR builds off of https://github.com/googleapis/python-bigtable/pull/924 to add instrumentation to each rpc in the Bigtable v3 data client

---

This change required some changes to the gapic generated client, which will need to be upstreamed to the gapic generator at some point: https://github.com/googleapis/gapic-generator-python/issues/1856

TODO: 
- ~~add end-to-end system tests~~
- get a final benchmarking in place
- improve _mutate_rows instrumentation
- more tests?
  - mutations batcher
  - _read_rows and _mutate_rows operations
    - sepcifically: _MutateRowsIncomplete exception


google/cloud/bigtable/data/_async/_mutate_rows.py
```diff
@@ -177,34 +193,47 @@ async def _run_attempt(self):
...
+                            # mutation failed; update error list (and remaining_indices if retryable)
+                            self._handle_entry_error(orig_idx, entry_error)
+                        elif orig_idx in self.errors:
+                            # mutation succeeded; remove from error list
+                            del self.errors[orig_idx]
+                        # remove processed entry from active list
+                        del active_request_indices[result.index]
+            finally:
+                # send trailing metadata to metrics
+                result_generator.cancel()
```
mutianf, line 216: why do we need to call cancel in the end?

google/cloud/bigtable/data/_async/_read_rows.py
```diff
@@ -144,70 +159,82 @@ def _read_rows_attempt(self) -> AsyncGenerator[Row, None]:
...
+                    if c.reset_row:
+                        current_key = None
+                    elif c.commit_row:
+                        # update row state after each commit
+                        self._last_yielded_row_key = current_key
+                        if self._remaining_count is not None:
+                            self._remaining_count -= 1
+                            if self._remaining_count < 0:
+                                raise InvalidChunk("emit count exceeds row limit")
+                        current_key = None
```
mutianf, line 210: Is it possible to refactor the code first and add the metric instrumentation in a separate PR? it's a bit hard to see what's actually changed.

google/cloud/bigtable/data/_async/_read_rows.py
```diff
@@ -284,7 +311,17 @@ async def merge_rows(
...
+                            # record first row latency in metrics
+                            is_first_row = False
+                            operation.attempt_first_response()
+                        block_time = time.monotonic()
                         yield Row(row_key, cells)
+                        # most metric operations use setters, but this one updates
+                        # the value directly to avoid extra overhead
+                        operation.active_attempt.application_blocking_time_ms += (  # type: ignore
+                            time.monotonic() - block_time
+                        ) * 1000
```
mutianf, line 324: Can we record everything in nanoseconds and convert them to milliseconds instead?

google/cloud/bigtable/data/_async/client.py
```diff
@@ -593,11 +595,20 @@ async def read_rows_stream(
         )
         retryable_excs = _get_retryable_errors(retryable_errors, self)
 
+        # extract metric operation if passed down through kwargs
+        # used so that read_row can disable is_streaming flag
+        metric_operation = kwargs.pop("metric_operation", None)
```
mutianf, line 600: Is this the API customer will interact with? This feels a bit weird 🤔  I think if someone calls readrows we can set is_streaming=True and OperationType to be ReadRows.

google/cloud/bigtable/data/_async/mutations_batcher.py
```diff
@@ -328,17 +331,26 @@ async def _flush_internal(self, new_entries: list[RowMutationEntry]):
         """
         # flush new entries
         in_process_requests: list[asyncio.Future[list[FailedMutationEntryError]]] = []
+        metric = self._table._metrics.create_operation(OperationType.BULK_MUTATE_ROWS)
```
mutianf, line 334: We don't have this OperationType in java 😓  let's just use MUTATE_ROWS for this.

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -395,7 +403,17 @@ def wrapped_predicate(exc: Exception) -> bool:
...
             return inner_result
 
-        return wrapped_predicate
+        def wrapped_exception_factory(
+            exc_list: list[Exception],
+            reason: RetryFailureReason,
+            timeout_val: float | None,
+        ) -> tuple[Exception, Exception | None]:
+            exc, source = _retry_exception_factory(exc_list, reason, timeout_val)
+            if reason != RetryFailureReason.NON_RETRYABLE_ERROR:
```
mutianf, line 412: should this be if reason = RetryFailureReason.NON_RETRYABLE_ERROR?

daniel-sanche: No, the `wrapped_predicate` will report all exceptions encountered. But the predicate isn't called when the operation ends due to timeout. So this extra wrapper is needed for that

This part is a bit more complicated than I hoped, so I was considering trying to refactor some of the read_rows instrumentation when I have a chance. Let me know what you think

google/cloud/bigtable/data/_async/_mutate_rows.py
daniel-sanche: This is calling cancel on the grpc stream. I think this is in case we encounter a client error; we need to stop the stream so it'll give us the trailing metadata

google/cloud/bigtable/data/_async/client.py
daniel-sanche: This is a hidden argument, users aren't supposed to interact with it

The problem is read_row is a small wrapper on top of read_rows. So read_rows can't assume the operation is streaming. This pattern is trying to allow read_row to pass down its own operation when it calls read_rows, so it can make sure streaming is False

We could also solve it using an entirely separate helper if this is too ugly though

google/cloud/bigtable/data/_async/mutations_batcher.py
daniel-sanche: BULK_MUTATE_ROWS is just the name we've been using for mutate_rows, since it's easier to parse. The string value is still "MutateRows"

google/cloud/bigtable/data/_async/client.py
mutianf: > users aren't supposed to interact with it

Is it possible to make sure users can't interact with it?

google/cloud/bigtable/data/_async/mutations_batcher.py
mutianf: Gotcha!

google/cloud/bigtable/data/_async/_mutate_rows.py
mutianf: What client error would happen? I'm just wondering if this will introduce some bugs (like not consuming all the results or cancelling the grpc stream twice)

google/cloud/bigtable/data/_async/_read_rows.py
```diff
@@ -144,70 +160,82 @@ def _read_rows_attempt(self) -> AsyncGenerator[Row, None]:
...
             return
         it = chunks.__aiter__()
+        is_first_row = True
         # For each row
         while True:
             try:
                 c = await it.__anext__()
             except StopAsyncIteration:
                 # stream complete
+                operation.end_with_success()
```
mutianf, line 238: does this happen when customer cancels the read in the middle of a stream?

google/cloud/bigtable/data/_metrics/data_model.py
mutianf: I see. It makes sense but a little hard to follow. Maybe add a comment for now?  And also in _read_rows start_operation  comment what `metric_fns[0]` and `metric_fns[1]` are?

google/cloud/bigtable/data/_metrics/handlers/opentelemetry.py
```diff
@@ -211,7 +212,8 @@ def on_attempt_complete(
             combined_throttling += op.flow_throttling_time_ms
         self.otel.throttling_latencies.record(combined_throttling, labels)
         self.otel.application_latencies.record(
-            attempt.application_blocking_time_ms + attempt.backoff_before_attempt_ms, labels
+            attempt.application_blocking_time_ms + attempt.backoff_before_attempt_ms,
+            labels,
         )
```
danieljbruce, line 217: Maybe this was added in an earlier PR, but first response latencies is actually per operation not per attempt.

daniel-sanche:
closing in favor of https://github.com/googleapis/python-bigtable/pull/1187

PR #947, opened 2024-03-20
chore(docs): added generated docs for async client

daniel-sanche:
Added autogenerated sphinx docs pages for async data client

Currently serving them in the Using the API section:

image

---

To view docs locally:
- run `nox -s docs` to generate html pages
- run `python -m http.server -d ./docs/_build/html` to start a local server on `localhost:8000`

BEGIN_COMMIT_OVERRIDE
chore(docs): added generated docs for async client
END_COMMIT_OVERRIDE

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -1,5 +1,5 @@
-Using the Standard Client
-=========================
+Using the Sync Client
+=====================
 
 .. toctree::
   :maxdepth: 2
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -1,5 +1,5 @@
-Using the API
-=============
+Using the Standard Client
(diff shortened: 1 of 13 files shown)


docs/usage.rst
```diff
@@ -1,5 +1,5 @@
-Using the API
-=============
+Using the Standard Client
```
mutianf, line 1: nit: standard client -> sync client?

mutianf (approved):

docs/usage.rst
daniel-sanche: Eventually there will be two sync clients, so we'll need a different naming system. I'll change it to sync for now though, we can discuss that later!

PR #961, opened 2024-05-15
doc: add samples for filtering using async apis

igorbernstein2:
This adds missing samples for filters when using the async api

### DIFF
--- a/samples/snippets/filters/filter_snippets_async.py
+++ b/samples/snippets/filters/filter_snippets_async.py
@@ -1,0 +1,335 @@
+# Copyright 2024, Google LLC
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


samples/snippets/filters/filter_snippets_async.py
```diff
@@ -0,0 +1,337 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import datetime
+from google.cloud.bigtable.data import Row
+from google.cloud._helpers import _datetime_from_microseconds
+
+
+# [START bigtable_filters_limit_row_sample]
```
daniel-sanche, line 19: Don't we need to add an `async` prefix to the region tags, to avoid clobbering the sync ones?

daniel-sanche (changes requested):

samples/snippets/filters/filter_snippets_async.py
igorbernstein2: I added asyncio :)

daniel-sanche (approved):

PR #962, opened 2024-05-15
chore(docs): improve devsite structure

daniel-sanche:
## Changes

Previously, the [devsite page](https://cloud.google.com/python/docs/reference/bigtable/latest) for bigtable was messy, combining pages for both clients into a single list:

image

This change adds a post-processing step to `docfx`, to add sections for the two client types to the table of contents, to keep things organized

image

---

## Implementation Details

The table of contents on devsite is configured through `docs/_build/html/docfx_yaml/toc.yml`, which is generated uisng `nox -s docfx` as part of the release pipeline.

This PR adds a new script called `patch_devsite_toc.py`, which is added as a post-processing step to the end of `nox -s docfx` to add new sections to the auto-generated table of contents.

As part of the PR, all pages specific to each client have been moved into subfolders in the `docs/` directory. These sub-folders can then be referenced by `patch_devsite_toc.py` when adding new sections to the table of contents

To avoid docs issues in the future, I also added a verification step to the new script, to make sure the table of contents remains formatted as expected

Staged page available to reviewers on request


docs/standard_client/usage.rst
```diff
@@ -1,10 +1,15 @@
-Using the Sync Client
-=====================
+Standard Client
```
daniel-sanche, line 1: Is "Standard Client" a good name for this?

dandhlee: Potentially? Not sure what's a good name for it though. `Bigtable client` might be a more generically encapsulating name, but might be confusing with the API section right below it

docs/scripts/patch_devsite_toc.py
```diff
@@ -0,0 +1,173 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+python-bigtable is made up of multiple clients, which we want to categorize separately in
+the devsite table of contents.
+
+This script allows us to add separate sections to the autogenerated toc.yml file to
+accomplish this.
```
igorbernstein2, line 19: Please add a bit more background here:

This script will run after `nox -s docfx` is run. `docfx` is the api doc format used by google cloud. It is described here: https://github.com/googleapis/docuploader?tab=readme-ov-file#requirements-for-docfx-yaml-tarballs. 
One of the file used by docfx is toc.yml which is used to generate the table of contents sidebar. This script will patch file to create subfolders for each of the clients

docs/scripts/patch_devsite_toc.py
```diff
@@ -0,0 +1,173 @@
...
+        :param index_file_name: name of an index file within dir_name. This file
+            will not be included in the table of contents, but provides an ordered
+            list of the other files which should be included
+        """
+        self.dir_name = dir_name
+        self.index_file_name = index_file_name
+        index_file_path = os.path.join(dir_name, index_file_name)
+        # find set of files referenced by the index file
+        with open(index_file_path, "r") as f:
+            self.title = f.readline().replace("\n", "")
```
igorbernstein2, line 67: nit - .strip()

docs/scripts/patch_devsite_toc.py
```diff
@@ -0,0 +1,173 @@
...
+            in_toc = False
+            self.items = []
+            for line in f:
```
igorbernstein2, line 70: I think this can be easier to read if flattened a bit:

```python
for line in f:
  # ignore empty lines
  if not line.rstrip('\n'):
     continue
  if line.startswith(".. toctree::"):
     in_toc = True
     continue
  # ignore directives
  if ":" in line:
     continue
  if not in_toc:
    continue
  # bail when toc indented block is done
  if not line.startswith(" ") or not line.startswith("\t"):
    # done with indented toc
    break
  # extract entries
  self.items.append(extract_toc_entry(line.strip())
```

docs/scripts/patch_devsite_toc.py
```diff
@@ -0,0 +1,173 @@
...
+        are not copied over by the docfx build by default
+        """
+        for file in os.listdir("_build/markdown/" + self.dir_name):
+            shutil.copy(
+                f"_build/markdown/{self.dir_name}/{file}",
+                f"_build/html/docfx_yaml",
+            )
+
+
+def validate_toc(toc_file_path, expected_section_list, added_sections):
```
igorbernstein2, line 113: at what point in the release process does this run? Can we end up in a state that the artifact was uploaded to pypi but this failed so we end up with a release egg but no docs?

If this is the case, should the validation happen as a presubmit instead?

docs/scripts/patch_devsite_toc.py
```diff
@@ -0,0 +1,173 @@
...
+                        # ignore empty lines
+                        continue
+                    elif line.startswith(" ") or line.startswith("\t"):
+                        # keep track of the items in the table of contents
+                        # ignore : directives
+                        if ":" not in line:
+                            file_name = line.strip()
+                            # load the file to get the title
+                            with open(f"{dir_name}/{file_name}.rst", "r") as f2:
+                                file_title = f2.readline().replace("\n", "")
```
igorbernstein2, line 84: nit rstrip("\n")

docs/scripts/patch_devsite_toc.py
```diff
@@ -0,0 +1,173 @@
...
+        """
+        Convert the TocSection object to a dictionary that can be written to a yaml file
+        """
+        return {"name": self.title, "items": self.items}
+
+    def copy_markdown(self):
+        """
+        Copy markdown files from _build/markdown/dir_name to _build/html/docfx_yaml
+
+        This is necessary because the markdown files in dub-directories
```
igorbernstein2, line 103: s/dub/sub

docs/standard_client/usage.rst
daniel-sanche: Maybe "Classic Client"? There's the complication that we also have to communicate the differences between the admin api vs the data api, so API is a bit overloaded

daniel-sanche:
Added a feature to remove sections, and used it to remove the bottom "Bigtable" section. This section was a duplicate of the content we have in the two client folders

docs/scripts/patch_devsite_toc.py
```diff
@@ -0,0 +1,201 @@
...
+    # save file
+    if output_file_path is None:
+        output_file_path = toc_file_path
+    with open(output_file_path, "w") as f:
+        yaml.dump(current_toc, f)
+
+
+def remove_sections(toc_file_path, section_list, output_file_path=None):
+    """
+    Add new sections to the autogenerated docfx table of contents file
```
igorbernstein2, line 59: I think the doc is a copy/paste error?

docs/standard_client/usage.rst
igorbernstein2: Classic client seems good

igorbernstein2 (approved):

PR #966, opened 2024-05-17
chore(docs): improve write_batch async sample

daniel-sanche:
- fix missing `await` in async write_batch sample
- add exception handling to sample, giving info on each underlying failure

### DIFF
--- a/samples/snippets/data_client/data_client_snippets_async.py
+++ b/samples/snippets/data_client/data_client_snippets_async.py
@@ -55,6 +55,8 @@ async def write_batch(table):
                             SetCell(family_id, "connected_wifi", 1),
                             SetCell(family_id, "os_build", "12155.0.0-rc1"),
                         ]
+                        # awaiting the batcher.append method adds the RowMutationEntry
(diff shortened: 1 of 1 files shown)


samples/snippets/data_client/data_client_snippets_async.py
```diff
@@ -42,24 +42,32 @@ async def write_batch(table):
...
-                        RowMutationEntry("tablet#a0b81f74#20190502", mutation_list)
-                    )
+                try:
+                    async with table.mutations_batcher() as batcher:
+                        mutation_list = [
+                            SetCell(family_id, "connected_cell", 1),
+                            SetCell(family_id, "connected_wifi", 1),
+                            SetCell(family_id, "os_build", "12155.0.0-rc1"),
+                        ]
+                        await batcher.append(
```
igorbernstein2, line 58: please add a comment that this await is for scheduling the mutation and not for applying the mutation...that will only happen when the context manager exits or batcher.close() is called

igorbernstein2 (approved):
lgtm after the comment is added

PR #970, opened 2024-05-28
fix: improve backup name regex when listing backups 

### LINKED ISSUE #969: Backup name format does not match the Bigtable Admin API
The regular expression used in [backup.py](https://github.com/googleapis/python-bigtable/blob/v2.23.1/google/cloud/bigtable/backup.py#L31) does not match the format specified in the [Bigtable Admin API](https://cloud.google.com/bigtable/docs/reference/admin/rest/v2/projects.instances.clusters.backups/create).

#### Environment details

  - OS type and version: Mac OC 13.6.6
  - Python version:  3.11.8
  - pip version: 22.3.1 
  - `google-cloud-bigtable` version: 2.22.0

#### Steps to reproduce

  1. Create a Bigtable table instance
  2. Call the list_backup method

#### Code example

```python
from google.cloud import bigtable

client = bigtable.Client(admin=True, project='XXXX')
instance = client.instance('XXXX')
instance.table('_locks_').list_backups()
```

#### Stack trace
```
Traceback (most recent call last):
  File "XXX/bigtable_regex_error.py", line 5, in <module>
    instance.table('_locks_').list_backups()
  File "XXX/venv/lib/python3.11/site-packages/google/cloud/bigtable/table.py", line 1000, in list_backups
    result.append(Backup.from_pb(backup_pb, self._instance))
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "XXX/venv/lib/python3.11/site-packages/google/cloud/bigtable/backup.py", line 253, in from_pb
    raise ValueError(
ValueError: ('Backup protobuf name was not in the expected format.', 'projects/XXXX/instances/XXXX/clusters/XXXX/backups/_locks_-backup-20240527-0700')
```

AmitMatsil:
Change the regular expression to match the format specified in the Bigtable Admin API documentation: https://cloud.google.com/bigtable/docs/reference/admin/rest/v2/projects.instances.clusters.backups/create

Fixes #969

### DIFF
--- a/google/cloud/bigtable/backup.py
+++ b/google/cloud/bigtable/backup.py
@@ -28,7 +28,7 @@ _BACKUP_NAME_RE = re.compile(
     r"^projects/(?P<project>[^/]+)/"
     r"instances/(?P<instance_id>[a-z][-a-z0-9]*)/"
     r"clusters/(?P<cluster_id>[a-z][-a-z0-9]*)/"
-    r"backups/(?P<backup_id>[a-z][a-z0-9_\-]*[a-z0-9])$"
+    r"backups/(?P<backup_id>[_a-zA-Z0-9][-_.a-zA-Z0-9]*)$"
 )
 
 _TABLE_NAME_RE = re.compile(


igorbernstein2 (approved):
Thanks for the fix!
I dont really think we should be doing these kind of validations on the client to begin with. 
I think I would prefer for the regex to just be something like: `backups/(?P<backup_id>[^/]+)` and just have the server do all of the validation.

This PR is an improvement so I'm ok merging it as is. However in the future we really should stop duplicating server logic here.

PR #977, opened 2024-06-04
feat: improve async sharding

daniel-sanche:
Use a semaphore to control shard concurrency, rather than discrete batches

### DIFF
--- a/google/cloud/bigtable/data/_async/client.py
+++ b/google/cloud/bigtable/data/_async/client.py
@@ -750,6 +750,7 @@ class TableAsync:
 
         # limit the number of concurrent requests using a semaphore
         concurrency_sem = asyncio.Semaphore(_CONCURRENCY_LIMIT)
+
         async def read_rows_with_semaphore(query):
             async with concurrency_sem:
                 # calculate new timeout based on time left in overall operation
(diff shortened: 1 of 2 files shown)


igorbernstein2 (approved):

google/cloud/bigtable/data/_async/client.py
```diff
@@ -755,6 +755,10 @@ async def read_rows_with_semaphore(query):
             async with concurrency_sem:
                 # calculate new timeout based on time left in overall operation
                 shard_timeout = next(rpc_timeout_generator)
+                if shard_timeout <= 0:
+                    raise DeadlineExceeded(
+                        "Operation timeout exceeded before starting query"
```
igorbernstein2, line 760: maybe before starting subquery?

igorbernstein2 (approved):

google/cloud/bigtable/data/_async/client.py
daniel-sanche: This is the earliest we can check after getting through the semaphore. 

There is a similar check at the very beginning as part of `_get_timeouts` though (which raises an `AttributeError`)

Issue #941, opened 2024-03-15
Support timeout / retry on Table.read_row()


mwarkentin:
**Is your feature request related to a problem? Please describe.**

We recently experienced an issue where a small number of requests to BigTable using [Table.read_row()](https://cloud.google.com/python/docs/reference/bigtable/latest/google.cloud.bigtable.table.Table#google_cloud_bigtable_table_Table_read_row) hung until we hit the default 60s timeout. This caused issues with our application including the kafka consumer being kicked out of the consumer group after not responding for 30s.

We want to set a more aggressive timeout for this function call (eg. less than 30 seconds) however we haven't been able to figure out a way - `read_row()` does not seem to support either `timeout` or `retry` args which are the documented ways to set a timeout on a request.

 **Describe the solution you'd like**

Support `table.read_row("keyid", timeout=5)`

Alternatively, support setting a new global default timeout via environment variable or config file.

 **Describe alternatives you've considered**

Perhaps we could refactor our code to use `read_rows()` instead and fetch a single row, but this feels hacky.

 **Additional context**

* Existing support issue: https://console.cloud.google.com/support/cases/detail/v2/49879020
* Google account team members: Jason Shin, Scott La Rue

mwarkentin:
@daniel-sanche @ayubun looks like this was tackled in https://github.com/googleapis/python-bigtable/pull/982 (THANKS!!)

Should I close this issue now? (I haven't had a chance to validate the update yet though)

mwarkentin:
Ah, I see that release is not actually available yet. Will wait to confirm. :)

daniel-sanche:
I just cut the release for `v2.27.0`, so I think we should be able to close this. Thanks!

PR #982, opened 2024-06-12
feat: surface `retry` param to `Table.read_row` api

ayubun:
### what ? (￣^￣ゞ
this pr surfaces a `retry` param for `Table.read_row`. this param defaults to `DEFAULT_RETRY_READ_ROWS` just as it does for `Table.read_rows`. it then passes forward into the `Table.read_rows` call

### why ? („• ᴗ •„)
in the case where a caller may want to override the default `retry` param for the `Table.read_row` api, they currently *must* switch to using `Table.read_rows` and add their own code for pulling the first item out of the iterator and guarding against multi-row responses (which are functionalities that `Table.read_row` already provides)

in an ideal world, the `Table.read_row` helper method can accept and pass along the retry param so that clients don't need to write their own duplicate implementations wrapping `Table.read_rows`~

### related issue ౨ৎ⋆˚｡⋆
implements #941

### DIFF
--- a/google/cloud/bigtable/table.py
+++ b/google/cloud/bigtable/table.py
@@ -533,7 +533,7 @@ class Table(object):
             for cluster_id, value_pb in table_pb.cluster_states.items()
         }
 
-    def read_row(self, row_key, filter_=None):
+    def read_row(self, row_key, filter_=None, retry=DEFAULT_RETRY_READ_ROWS):
         """Read a single row from this table.
 
         For example:
@@ -550,6 +550,14 @@ class Table(object):
         :param filter_: (Optional) The filter to apply to the contents of the
                         row. If unset, returns the entire row.
 
+        :type retry: :class:`~google.api_core.retry.Retry`
+        :param retry:
+            (Optional) Retry delay and deadline arguments. To override, the
+            default value :attr:`DEFAULT_RETRY_READ_ROWS` can be used and
+            modified with the :meth:`~google.api_core.retry.Retry.with_delay`
+            method or the :meth:`~google.api_core.retry.Retry.with_deadline`
+            method.
+
         :rtype: :class:`.PartialRowData`, :data:`NoneType <types.NoneType>`
         :returns: The contents of the row if any chunks were returned in
                   the response, otherwise :data:`None`.
@@ -558,7 +566,9 @@ class Table(object):
         """
         row_set = RowSet()
         row_set.add_row_key(row_key)
-        result_iter = iter(self.read_rows(filter_=filter_, row_set=row_set))
+        result_iter = iter(
+            self.read_rows(filter_=filter_, row_set=row_set, retry=retry)
+        )
         row = next(result_iter, None)
         if next(result_iter, None) is not None:
             raise ValueError("More than one row was returned.")


ayubun:
i wanted to bump this since it's been a while ^^; @rkaregar

daniel-sanche (approved):

google/cloud/bigtable/table.py
```diff
@@ -558,7 +566,7 @@ def read_row(self, row_key, filter_=None):
         """
         row_set = RowSet()
         row_set.add_row_key(row_key)
-        result_iter = iter(self.read_rows(filter_=filter_, row_set=row_set))
+        result_iter = iter(self.read_rows(filter_=filter_, row_set=row_set, retry=retry))
```
daniel-sanche, line 569: Thanks for submitting this

I can't merge due to a lint error. i think it's because this line is too long. Maybe break it up:

```
        result_iter = iter(
            self.read_rows(filter_=filter_, row_set=row_set, retry=retry)
        )

```

ayubun: @daniel-sanche done https://github.com/googleapis/python-bigtable/commit/39e8f90b12ab14cd7981bf7f1f1d80caa9a1447d

daniel-sanche (approved):
There seems to be an issue in the owlbot test, unreleated to these changes. I'll merge this when that is resolved

daniel-sanche:
Actually, it looks like it's failing on the [noxfile customizations](https://github.com/ayubun/python-bigtable/blob/39e8f90b12ab14cd7981bf7f1f1d80caa9a1447d/owlbot.py#L101). These were stripped out in the [upstream main owlbot.py file](https://github.com/googleapis/python-bigtable/blob/main/owlbot.py). So I think the issue is because the tests are running against your branch, which doesn't contain all the upstream changes. 

Please merge in upstream's main, and then we should be good to go

ayubun:
@daniel-sanche done

daniel-sanche:
thanks! There's now a CI issue on the backend, but I will merge this when that is resolved

mwarkentin:
@ayubun @daniel-sanche #941 mentioned both retries and timeouts - I'm not sure but it looks like this adds retry but not timeout config? Or are they combined somehow in the `retry` param?

mwarkentin:
Ah, I see: https://github.com/ayubun/python-bigtable/blob/c8c04d35b6e236fe98d350ca45dc49fc059da8cb/google/cloud/bigtable/row_data.py#L96

mwarkentin:
Does this look like the right way to override the timeout?

```python
from google.cloud.bigtable.row_data import DEFAULT_RETRY_READ_ROWS

modified_retry = DEFAULT_RETRY.with_deadline(5.0)
row = self._get_table().read_row(key, retry=modified_retry)
```

daniel-sanche:
@mwarkentin Yeah, that code you supplied looks right. Technically you should be using [`.with_timeout`](https://github.com/googleapis/python-api-core/blob/66d84a3a9ef0ab0c2124a1bf12cc3c45c2393e74/google/api_core/retry/retry_base.py#L292) ([explaination here](https://github.com/googleapis/python-api-core/blob/66d84a3a9ef0ab0c2124a1bf12cc3c45c2393e74/google/api_core/retry/retry_unary.py#L195)), but deadline is still supported as an equivalent alias

This uses the api_core.Retry class, which is shared across all GCP python libraries for retry/timeout/backoff configuration. You can find more [documentation here](https://github.com/googleapis/python-api-core/blob/66d84a3a9ef0ab0c2124a1bf12cc3c45c2393e74/google/api_core/retry/retry_unary.py#L85)

mwarkentin:
Thanks!

daniel-sanche:
Thanks, I opened a request to have that fixed!

PR #986, opened 2024-07-05
chore: update templated files

parthea:


daniel-sanche (changes requested):
looks like this is removing some important owlbot customizations. I'll push a fix

parthea:
Good catch ! I see that now

noxfile.py
```diff
@@ -425,15 +436,19 @@ def docfx(session):
         os.path.join("docs", ""),
         os.path.join("docs", "_build", "html", ""),
     )
-    # Customization: Add extra sections to the table of contents for the Classic vs Async clients
-    session.install("pyyaml")
-    session.run("python", "docs/scripts/patch_devsite_toc.py")
```
parthea, line 430: This customization should be restored

daniel-sanche (approved):

PR #988, opened 2024-07-07
fix: update FLUSH_COUNT and MAX_MUTATION_SIZE to align with documentation

nguymin4:
This PR fix discrepancies between documentation and actual values of `FLUSH_COUNT` and `MAX_MUTATION_SIZE`

Of course we can instead update the documentation. However I do believe that `FLUSH_COUNT = 100` is quite low and doesn't have the same magnitude compared to `MAX_MUTATION_SIZE = 20MB`

### DIFF
--- a/google/cloud/bigtable/batcher.py
+++ b/google/cloud/bigtable/batcher.py
@@ -25,7 +25,7 @@ from dataclasses import dataclass
 
 FLUSH_COUNT = 1000  # after this many elements, send out the batch
 
-MAX_MUTATION_SIZE = 20 * 1024 * 1024  # 20MB # after this many bytes, send out the batch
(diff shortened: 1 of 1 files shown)


daniel-sanche:
I believe the max bytes should be 20mb, but the count limit should be 1000.

That said, we're working on replacing this batcher with a new version, so I'd rather not change these parameters on this one for now

Where in the docs were you seeing the discrepency?

nguymin4:
@daniel-sanche 
In this same file, the function documentation says max count is 1000 and max bytes is 5 MB

2024-10-26T14:59:43,325096688+03:00

But good to know that you're working on new version of this. Because current, I'm using it via Apache Beam (Dataflow) and the performance is really bad with python-bigtable 2.x.

PR #1012, opened 2024-08-08
docs: add clarification around SQL timestamps

jackdingilian:

### DIFF
--- a/google/cloud/bigtable/data/execute_query/metadata.py
+++ b/google/cloud/bigtable/data/execute_query/metadata.py
@@ -214,6 +214,12 @@ class SqlType:
         type_field_name = "bool_type"
 
     class Timestamp(Type):
+        """
+        Timestamp supports :class:`DatetimeWithNanoseconds` but Bigtable SQL does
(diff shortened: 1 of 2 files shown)


daniel-sanche (approved):

google/cloud/bigtable/data/execute_query/metadata.py
```diff
@@ -214,6 +214,11 @@ class Bool(Type):
         type_field_name = "bool_type"
 
     class Timestamp(Type):
+        """
+            Timestamp supports :class:`DatetimeWithNanoseconds` but Bigtable SQL does
+            not currently support nanoseconds precision. We support this for potential
+            compatibility in the future. Nanoseconds are currently ignored.
+        """
```
daniel-sanche, line 221: It looks like this is what needs to change:

```suggestion
        """
        Timestamp supports :class:`DatetimeWithNanoseconds` but Bigtable SQL does
        not currently support nanoseconds precision. We support this for potential
        compatibility in the future. Nanoseconds are currently ignored.
        """
```

PR #1020, opened 2024-09-04
chore(docs): use read_rows_stream in filter samples

daniel-sanche:
replace read_rows with read_rows_stream, as it is more efficient


igorbernstein2:
what makes this more efficient?

daniel-sanche:
> what makes this more efficient?

context: b/347322404

read_rows buffers the entire output before returning the resulting list, read_rows_stream streams gives a generator for acting on the results as they arrive

daniel-sanche:
After discussing this offline, I think we will leave as-is

PR #1034, opened 2024-10-31
feat: support authorized views

daniel-sanche:
This PR adds support for [authorized views](https://cloud.google.com/bigtable/docs/authorized-views):

> Authorized views are logical views of tables that give you fine-grained access control of your Bigtable data. An authorized view is a subset of a table that you configure to include specific table data. Then you grant access to the authorized view separately from access to the table.

To implement this feature, we added a new `_ApiSurface` super class for the `Table` object, and moved all logic into the super class. Then, we added a new subclass for `AuthorizedView`, alongside the existing `Table`. Authorized views should be functionally identical to tables from the client-library level, but they can be configured to only read/write to sepecific regions of the database


google/cloud/bigtable/data/_async/client.py
```diff
@@ -510,6 +520,72 @@ def get_table(self, instance_id: str, table_id: str, *args, **kwargs) -> TableAs
...
+                "",
+            ),
+            "RAISE_NO_LOOP": (
+                "RuntimeError: if called outside of an async context (no running event loop)",
+                "None",
+            ),
+        },
+    )
+    def get_authorized_view(
+        self, instance_id: str, table_id: str, view_id: str, *args, **kwargs
```
andre-sampaio, line 537: `:%s/view_id/authorized_view_id`. I'd rather be very explicit on public APIs so users get used to being explicit about views as well (in the light of materialized views)

google/cloud/bigtable/data/_async/client.py
```diff
@@ -612,13 +688,12 @@ async def __aexit__(self, exc_type, exc_val, exc_tb):
         await self._gapic_client.__aexit__(exc_type, exc_val, exc_tb)
 
 
-@CrossSync.convert_class(sync_name="Table", add_mapping_for_name="Table")
-class TableAsync:
+@CrossSync.convert_class(sync_name="_ApiSurface")
+class _ApiSurfaceAsync:
```
andre-sampaio, line 692: NIT: Maybe `_DataApiTarget` or something like this? Not something I'm particularly worried about since it is not part of the public API, so feel free to ignore this comment

google/cloud/bigtable/data/_async/client.py
```diff
@@ -742,6 +817,9 @@ def __init__(
             default_mutate_rows_retryable_errors or ()
         )
         self.default_retryable_errors = default_retryable_errors or ()
+        # used to populate table_name or authorized_view_name fields in request protos
+        self._request_path = {"table_name": self.table_name}
```
andre-sampaio, line 821: Can we move this + 

```
self.table_id = table_id
        self.table_name = self.client._gapic_client.table_path(
            self.client.project, instance_id, table_id
        )
```

to the Table class?

google/cloud/bigtable/data/_async/_mutate_rows.py
```diff
@@ -70,7 +74,7 @@ class _MutateRowsOperationAsync:
     def __init__(
         self,
         gapic_client: GapicClientType,
-        table: TableType,
+        table: ApiSurfaceType,
```
andre-sampaio, line 77: NIT: `:%s/table/target` (or something else that indicates this isn't necessarily a table)

andre-sampaio: Same comment for other lines we have changed the type from `Table` to `ApiSurfaceType`

google/cloud/bigtable/data/_async/client.py
daniel-sanche: Sure, I like that better. Fixed

google/cloud/bigtable/data/_async/_mutate_rows.py
daniel-sanche: That's doable for the internal functions, but it could technically be a breaking change for MutationsBatcher. Do you think we can keep calling it `table` there for now, and open a bug to fix it in the major launch we have planned for this year?

Most people would probably use `client.mutations_batcher()` or `MutationsBatcher(my_target)`, but if they currently use `MutationsBatcher(table=my_target)`, that would be a break. And I don't think we can make it backwards compatible, since we set table as a required arg

google/cloud/bigtable/data/_async/client.py
daniel-sanche: I can make the request_path parts a bit cleaner. But don't authorized views also need to contain a table_id and table_name? Or am I misunderstanding the structure here?

andre-sampaio: Ah that's right, I was thinking of Materialized views (which don't have a table id). I'm good with keeping this here (though we may need to review it when we implement materialized view but that's for another time)

google/cloud/bigtable/data/_async/_mutate_rows.py
andre-sampaio: yup! I'm good with that since we have a clear upgrade path

andre-sampaio (approved):

PR #1035, opened 2024-11-01
chore: remove pooled transport

daniel-sanche:
Remove grpc channel pooling functionality.

After some benchmarking, we found pooling provided little benefit in python, and it added significant complexity

Also removing the two submodules from this repo, as we no longer have any gapic changes that need to be pushed upstream


google/cloud/bigtable/data/_async/client.py
```diff
@@ -131,12 +127,9 @@ def __init__(
                 on the client. API Endpoint should be set through client_options.
         Raises:
             RuntimeError: if called outside of an async context (no running event loop)
-            ValueError: if pool_size is less than 1
         """
-        # set up transport in registry
-        transport_str = f"pooled_grpc_asyncio_{pool_size}"
-        transport = PooledBigtableGrpcAsyncIOTransport.with_fixed_size(pool_size)
-        BigtableClientMeta._transport_registry[transport_str] = transport
+        if "pool_size" in kwargs:
```
mutianf, line 131: Could someone create a client with BigtableDataClientAsync("project-id", 3)?

mutianf (approved):

google/cloud/bigtable/data/_async/client.py
daniel-sanche: we discussed this offline, but the `__init__()` function includes a `*`, meaning only key-word arguments are allowed

PR #1043, opened 2024-11-15
chore: update post processor

parthea:
The change in `.github/release-trigger.yml` is needed to ensure that releases are triggered properly.


daniel-sanche:
It looks like I just merged these changes through https://github.com/googleapis/python-bigtable/pull/1052. Updating against main resulted in an empty diff. But let me know if there were important changes that still need to get in

PR #1075, opened 2025-01-17
feat: Add support for array and float32 SQL query params

jackdingilian:
One note: we currently support inferring the type of simple query params. Here I opt not to attempt this for lists. We could infer the type from the first element of the list (and error if it is None, or infer from the first non-null element). For now I think this approach is less error prone, and we can ease this restriction in the future


google/cloud/bigtable/data/execute_query/_parameters_formatting.py
```diff
@@ -109,6 +110,16 @@ def _detect_type(value: ExecuteQueryValueType) -> SqlType.Type:
             "Cannot infer type of None, please provide the type manually."
         )
 
+    if isinstance(value, list):
+        raise ParameterTypeInferenceFailed(
```
daniel-sanche, line 114: Can you declare this in the [execute_query docstring](https://github.com/googleapis/python-bigtable/blob/df08dcb553d354c782ebdf036f82bc1a77daf7c1/google/cloud/bigtable/data/_async/client.py#L567)? It looks like that's the public entrypoint for this

tests/system/data/test_system_async.py
```diff
@@ -14,13 +14,16 @@
 
 import pytest
```
daniel-sanche, line 15: We added [CrossSync](go/cross_sync_design) to this repo to generate the sync code from the async base. Since you're touching the async files, you''ll have to run `nox -s generate_sync` before merging (and a unit test will fail if the sync client isn't up to date, so you'll be reminded)

In more complex code, you sometimes need to add CrossSync annotations to the code to guide the generator, but you shouldn't have to do anything for the changes here, outside what you already have. But let me know if you run into issues with it

google/cloud/bigtable/data/execute_query/_parameters_formatting.py
jackdingilian: Done

tests/system/data/test_system_async.py
jackdingilian: Done, thanks

jackdingilian: Sorry I realized there was still something failing from the generated code. Regenerated and fixed now

daniel-sanche (approved):
LGTM

daniel-sanche:
Actually, it looks like Owlbot can't run on this PR since it's based on a fork. I remember seeing something about that being a new security policy. Could you re-submit this as an internal branch? (Let me know if you're missing permissions)

PR #1078, opened 2025-01-31
feat: Add support for array and float32 SQL query params

jackdingilian:
One note: we currently support inferring the type of simple query params. Here I opt not to attempt this for lists. We could infer the type from the first element of the list (and error if it is None, or infer from the first non-null element). For now I think this approach is less error prone, and we can ease this restriction in the future


daniel-sanche (approved):
LGTM

PR #1096, opened 2025-03-13
feat: Implement updated execute query protocol

jackdingilian:

### DIFF
--- a/google/cloud/bigtable/data/execute_query/_async/execute_query_iterator.py
+++ b/google/cloud/bigtable/data/execute_query/_async/execute_query_iterator.py
@@ -91,6 +91,9 @@ class ExecuteQueryIteratorAsync:
         """
         Collects responses from ExecuteQuery requests and parses them into QueryResultRows.
 
+        **Please Note** this is not meant to be constructed directly by applications. It should always
+        be created via the client. The constructor is subject to change.
+
         It is **not thread-safe**. It should not be used by multiple {TASK_OR_THREAD}.
 
         Args:
@@ -106,6 +109,7 @@ class ExecuteQueryIteratorAsync:
             retryable_excs: a list of errors that will be retried if encountered.
         Raises:
             {NO_LOOP}
+            :class:`ValueError <exceptions.ValueError>` as a safeguard if data is processed in an unexpected state
         """
         self._table_name = None
         self._app_profile_id = app_profile_id
@@ -217,6 +221,11 @@ class ExecuteQueryIteratorAsync:
 
     @CrossSync.convert(sync_name="__next__", replace_symbols={"__anext__": "__next__"})
     async def __anext__(self) -> QueryResultRow:
+        """
+        Yields QueryResultRows representing the results of the query.
+
+        :raises: :class:`ValueError <exceptions.ValueError>` as a safeguard if data is processed in an unexpected state
+        """
         if self._is_closed:
             raise CrossSync.StopIteration
         return await self._result_generator.__anext__()
--- a/google/cloud/bigtable/data/execute_query/_checksum.py
+++ b/google/cloud/bigtable/data/execute_query/_checksum.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8 -*-
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+import warnings
+
+with warnings.catch_warnings(record=True) as import_warning:
+    import google_crc32c  # type: ignore
+
+
+class _CRC32C(object):
+    """
+    Wrapper around ``google_crc32c`` library
+    """
+
+    warn_emitted = False
+
+    @classmethod
+    def checksum(cls, val: bytearray) -> int:
+        """
+        Returns the crc32c checksum of the data.
+        """
+        if import_warning and not cls.warn_emitted:
+            cls.warn_emitted = True
+            warnings.warn(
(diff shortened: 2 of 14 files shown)


google/cloud/bigtable/data/execute_query/_checksum.py
```diff
@@ -0,0 +1,52 @@
...
+            self._google_crc32c = google_crc32c
+            self.import_failed = False
+        except ImportError:
+            self.import_failed = True
+            warnings.warn(
+                "Unable to import 'google-crc32 library on this python version and architecture."
+                " Disabling ExecuteQuery checksum validation. Please upgrade to a supported version"
+                " and architecture if possible",
+                category=RuntimeWarning,
+            )
```
daniel-sanche, line 38: How often do you expect this feature to be used? It might be better to only show this if a checksum is encountered? Or execute_query is used?

daniel-sanche (changes requested):

setup.py
```diff
@@ -44,6 +44,7 @@
     "proto-plus >= 1.22.3, <2.0.0dev",
     "proto-plus >= 1.25.0, <2.0.0dev; python_version>='3.13'",
     "protobuf>=3.20.2,<6.0.0dev,!=4.21.0,!=4.21.1,!=4.21.2,!=4.21.3,!=4.21.4,!=4.21.5",
+    "google-crc32c>=1.5.0, <2.0.0dev",
```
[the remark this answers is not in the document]
jackdingilian: (Based on discussion offline) there is a python fallback built into this. I wrapped the import in a custom warning when it uses the fallback now

google/cloud/bigtable/data/execute_query/_checksum.py
```diff
@@ -0,0 +1,52 @@
...
+        """
+        return not self.import_failed
+
+    def checksum(self, val: bytearray) -> int:
+        """
+        Returns the crc32c checksum of the data. ``enabled`` should always be checked before calling.
+        """
+        # google_crc32c wants bytes. Use a memoryview to avoid a copy
+        memory_view = memoryview(val)
+        return self._google_crc32c.value(bytes(memory_view))
```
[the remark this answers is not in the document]
jackdingilian: Refactored this to be a simple static method now

google/cloud/bigtable/data/execute_query/_checksum.py
jackdingilian: Done

google/cloud/bigtable/data/execute_query/_async/execute_query_iterator.py
```diff
@@ -83,6 +84,7 @@ def __init__(
         app_profile_id: Optional[str],
         request_body: Dict[str, Any],
         prepare_metadata: Metadata,
+        crc32c_implementation: _CRC32C,
```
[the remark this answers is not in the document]
jackdingilian: This is removed but it raises an interesting question. I dont think we really want the constructor to be public here, as it will likely need to change when we add PreparedStatement. 

Maybe we should only expose an `ExecuteQueryIterator` interface publicly and move the details to an ExecuteQueryIteratorImpl? We can still break the interface currently bc of the preview warning but don't want to box us in after launch

google/cloud/bigtable/data/execute_query/_checksum.py
```diff
@@ -0,0 +1,43 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+import warnings
+
+with warnings.catch_warnings(record=True) as import_warning:
+    import google_crc32c  # type: ignore
+
+
+class _CRC32C(object):
```
daniel-sanche, line 22: It would be good to have a test for this class. Especially checking if the warning is emitted as expected

tests/unit/data/_async/test_client.py
```diff
@@ -36,11 +36,15 @@
 
 from google.cloud.bigtable.data._cross_sync import CrossSync
 from tests.unit.data.execute_query.sql_helpers import (
+    chunked_responses,
```
daniel-sanche, line 39: are there any system tests that can be added?

daniel-sanche (approved):
added a couple questions about tests, but LGTM

google/cloud/bigtable/data/execute_query/_async/execute_query_iterator.py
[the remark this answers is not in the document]
daniel-sanche: I think the interface/implementation option could work. We could also raise a warning/exception in the __init__, that we suppress when called internally (maybe by passing an extra _ignore_warnings init argument?). Or even just adding a warning to the __init__ docstring should be enough. 

I think any of those options should be enough to make changes freely in the future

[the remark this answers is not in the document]
jackdingilian: Ok cool, I added a docstring as that seems simplest and clear enough for users

google/cloud/bigtable/data/execute_query/_checksum.py
jackdingilian: added in checksum_test

tests/unit/data/_async/test_client.py
jackdingilian: I have a follow on with a couple of new integration tests. W.r.t the chunked responses there isn't a great way to verify if the backend is sending multiple chunks or a single chunk.

I do have a significant number of conformance tests ready to go for this logic though, and im planning on adding the testproxy support for python as a fast follow

tests/unit/data/execute_query/test_checksum.py
```diff
@@ -0,0 +1,35 @@
...
+        with warnings.catch_warnings(record=True) as import_warning:
+            from google.cloud.bigtable.data.execute_query._checksum import _CRC32C
+
+            assert import_warning == []
+            with warnings.catch_warnings(record=True) as first_call_warning:
+                assert _CRC32C.checksum(b"test") == 2258662080
+                assert (
+                    "Using pure python implementation of `google-crc32` for ExecuteQuery response validation"
+                    in str(first_call_warning[0])
+                )
```
daniel-sanche, line 35: nit: it would be good to make sure repeated calls only send one warning. And maybe also a test where no warnings are needed

jackdingilian: Done

PR #1099, opened 2025-03-18
feat: Refactor Metadata, add system tests, remove preview warning

jackdingilian:
Removes ProtoMetadata because all of it's functionality is generic and ok to expose to users (column names and types). The original reason for this was to support other schema types in the future, while hiding them from users. The metadata will always have columns and types though, so this isn't currently necessary


daniel-sanche (approved):
LGTM

PR #1142, opened 2025-06-23
feat: Async consistency polling harness

gkevinzheng:
This PR implements both the async client's `wait_for_consistency` and `__init__` methods, as well as the async polling harness class.

### DIFF
--- a/google/cloud/bigtable/admin_v2/overlay/services/bigtable_table_admin/async_client.py
+++ b/google/cloud/bigtable/admin_v2/overlay/services/bigtable_table_admin/async_client.py
@@ -50,7 +50,7 @@ from google.cloud.bigtable.admin_v2.services.bigtable_table_admin import (
 from google.cloud.bigtable.admin_v2.services.bigtable_table_admin.transports.base import (
     BigtableTableAdminTransport,
 )
-from google.cloud.bigtable.admin_v2.overlay.types import async_consistency
+from google.cloud.bigtable.admin_v2.overlay.types import async_consistency, wait_for_consistency_request
 
 from google.cloud.bigtable.gapic_version import __version__ as bigtable_version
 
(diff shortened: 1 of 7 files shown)


google/cloud/bigtable/admin_v2/overlay/services/bigtable_table_admin/async_client.py
```diff
@@ -0,0 +1,228 @@
...
+        request: Optional[
+            Union[bigtable_table_admin.CheckConsistencyRequest, dict]
+        ] = None,
+        *,
+        name: Optional[str] = None,
+        consistency_token: Optional[str] = None,
+        retry: OptionalRetry = gapic_v1.method.DEFAULT,
+        timeout: Union[float, object] = gapic_v1.method.DEFAULT,
+        metadata: Sequence[Tuple[str, Union[str, bytes]]] = (),
+    ) -> async_consistency.AsyncCheckConsistencyPollingFuture:
```
daniel-sanche, line 144: I wonder if we want to return this custom Future directly. We could encapsulate it and use pure asyncio-native coroutines instead

That would be more idiomatic, but wouldn't be as consistent between sync/async. Do you think users will do anything with this future other than call `.result()`?

google/cloud/bigtable/admin_v2/overlay/services/bigtable_table_admin/async_client.py
```diff
@@ -0,0 +1,280 @@
...
+import functools
+
+from typing import Callable, Optional, Sequence, Tuple, Union
+from google.api_core import gapic_v1
+from google.api_core import retry as retries
+
+try:
+    OptionalRetry = Union[retries.Retry, gapic_v1.method._MethodDefault, None]
+except AttributeError:  # pragma: NO COVER
+    OptionalRetry = Union[retries.Retry, object, None]  # type: ignore
```
daniel-sanche, line 40: This should be AsyncRetry for the async client

daniel-sanche (changes requested):

daniel-sanche (approved):

PR #1144, opened 2025-06-25
feat: Reworked the wait_for_consistency call

gkevinzheng:
Changes:

1. Added a new wrapper class `WaitForConsistencyRequest` as an input parameter for `wait_for_consistency`. The wrapper class is proto plus based to be consistent with other GAPIC calls.
2. `wait_for_consistency` now generates a consistency token via `generate_consistency_token` then blocks on the polling harness, rather than returning the polling harness directly. This is consistent with the way Java does things. 
3. Because the polling harness is now not exposed to the user, it can be a private type.
4. Removed `wait_for_replication`.


daniel-sanche (approved):

google/cloud/bigtable/admin_v2/overlay/services/bigtable_table_admin/client.py
```diff
@@ -301,120 +291,74 @@ def sample_wait_for_consistency():
                 be of type `bytes`.
 
         Returns:
-            google.cloud.bigtable.admin_v2.overlay.types.CheckConsistencyPollingFuture:
-                An object representing a polling future.
-
-                The result type for the operation will be `bool`, and will return True when the
-                consistency check involving the given consistency token returns True.
+            bool:
```
mutianf, line 294: This only describes the standard_read_remote_mode. Can you also add the description of data_boost_read_local_writes here? Reference: https://source.corp.google.com/piper///depot/google3/google/bigtable/admin/v2/bigtable_table_admin.proto;l=1205?q=GEnerateConsistencyToken&sq=package:piper%20file:%2F%2Fdepot%2Fgoogle3%20-file:google3%2Fexperimental

google/cloud/bigtable/admin_v2/overlay/services/bigtable_table_admin/client.py
```diff
@@ -227,20 +231,18 @@ def sample_restore_table():
...
         metadata: Sequence[Tuple[str, Union[str, bytes]]] = (),
-    ) -> consistency.CheckConsistencyPollingFuture:
-        r"""Creates a polling future that periodically checks replication
-        consistency based on a consistency token, that is, if replication
-        has caught up based on the conditions specified in the token and
-        the check request. The future will stop checking once the underlying
-        :meth:`check_consistency` request involving that token returns True.
+    ) -> bool:
+        r"""Blocks until the mutations for the specified Table that have been
+        made before the call have been replicated. This is done by generating
```
mutianf, line 243:
```suggestion
        made before the call have been replicated or reads using an app profile with `DataBoostIsolationReadOnly`
        can see all writes committed before the token was created. This is done by generating
```

mutianf (approved):

PR #1151, opened 2025-07-15
chore(tests): system tests for autogen API

gkevinzheng:
Added system tests for the autogenerated Admin API. While working on the system tests, a number of issues popped up:

1. `WaitForConsistencyRequest` and `AsyncRestoreTableOperation` were not aliased under the `google.cloud.bigtable.admin_v2` package, so they were added to the appropriate `__init__.py` files.
2. The polling timeout for the `wait_for_consistency` polling harness has been extended to an indefinite timeout.
3. Updated the version of `google-api-core` to facilitate that

Test results:
```
$ nox -s system-3.8 -- -x
nox > Running session system-3.8
nox > Creating virtual environment (virtualenv) using python3.8 in .nox/system-3-8
nox > py.test --quiet --junitxml=system_3.8_sponge_log.xml tests/system -x
......sssssssssssssssssssssssssssssssssssssssssssssssssssssss............................... [ 37%]
............................................................................................ [ 74%]
.......................................s...................s..                               [100%]
========================================= warnings summary =========================================
nox > Session system-3.8 was successful.
```


tests/system/admin_overlay/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+INITIAL_CELL_VALUE = "Hello"
+NEW_CELL_VALUE = "World"
+
+
+@pytest.fixture(scope="session")
+def admin_overlay_project_id():
+    # Instantiate a meaningless client to get project name
+    # from environment variables.
+    client = ClientWithProject()
+    yield client.project
```
daniel-sanche, line 26: Creating a client just for this seems a bit messy. 

Looking through the ClientWithProject code, [it seems like it just uses](https://github.com/googleapis/python-cloud-core/blob/78ce8a6f7c781cf2585131d7961beec9069a206c/google/cloud/_helpers/__init__.py#L152C22-L152C43) `google.auth.default()` for this

I think you should also check `os.getenv("GOOGLE_CLOUD_PROJECT")` first though, so we can configure a different project if needed (And maybe put this up with the constants as ADMIN_TEST_PROJECT or something so it's discoverable?)

tests/system/admin_overlay/test_system_async.py
```diff
@@ -0,0 +1,364 @@
...
+def event_loop():
+    loop = asyncio.new_event_loop()
+    asyncio.set_event_loop(loop)
+    yield loop
+    loop.stop()
+    loop.close()
+
+
+@CrossSync.convert
+@CrossSync.pytest_fixture(scope="session")
```
daniel-sanche, line 63: This is making me realize @CrossSync.pytest_fixture should always do @CrossSync.convert automatically. We should change that later

tests/system/admin_overlay/test_system_async.py
```diff
@@ -0,0 +1,364 @@
...
+async def instances_to_delete(instance_admin_client):
+    instances = []
+
+    yield instances
+
+    for instance in instances:
+        await instance_admin_client.delete_instance(name=instance.name)
```
daniel-sanche, line 98: IIRC I think you might need a try/finally block here (and delete backups), to make sure these are always cleaned up even if there's an exception

```
async def instances_to_delete(instance_admin_client):
    instances = []
    try:
        yield instances
    finally:
        for instance in instances:
            await instance_admin_client.delete_instance(name=instance.name)
```

tests/system/admin_overlay/test_system_async.py
```diff
@@ -0,0 +1,364 @@
...
+            )
+        )
+
+    await data_client_table.bulk_mutate_rows(row_mutation_entries)
+
+
+@CrossSync.convert
+async def create_backup(
+    instance_admin_client, table_admin_client, instance, table, backups_to_delete
+) -> admin_v2.Backup:
```
daniel-sanche, line 204: docstring? It would be good to know why this helper is needed

tests/system/admin_overlay/test_system_async.py
```diff
@@ -0,0 +1,364 @@
...
+    await assert_table_cell_value_equal_to(
+        table_admin_client,
+        data_client,
+        instance_to_restore,
+        restored_table,
+        INITIAL_CELL_VALUE,
+    )
+
+
+@CrossSync.convert
```
daniel-sanche, line 333: nit: I think @CrossSync.pytest already does convert automatically, so this should be redundant. but you can double-check

daniel-sanche (commented):
a couple small comments, but looks good!

PR #1169, opened 2025-07-21
chore: rename MutationBatcher table argument to target

daniel-sanche:
When [Authorized View support ](https://github.com/googleapis/python-bigtable/pull/1034) was added, we updated `table` arguments to `target`, since views could be used in place of conventional table objects. But we couldn't make that rename to the MutationBatcher class, because it is publicly exposed and could be a breaking change.

This PR makes the change for inclusion in the next major release, `v3.0.0`

This PR also [starts an UPDATING.md migration guide](https://github.com/googleapis/google-cloud-python/issues/15297) to detail any breaking changes for the `v3.0.0` update. We can iterate on its contents over time

Fixes: https://github.com/googleapis/google-cloud-python/issues/15301


gkevinzheng (approved):

UPGRADING.md
```diff
@@ -0,0 +1,18 @@
...
+  - `google.cloud.bigtable.admin.BigtableTableAdminClient`
+  - `google.cloud.bigtable.admin.BigtableTableAdminClientAsync`
+
+The deprecated client will remain available as an alternative API surface, which internally delegates calls to the respective new clients. For most users, existing code will continue to work as before. But there may be some breaking changes associated with this update, which are detailed in this document.
+
+If you experience technical issues or have questions, please file an [issue](https://github.com/googleapis/python-bigtable/issues).
+
+## Breaking Changes
+- **[MutationBatcher](https://github.com/googleapis/python-bigtable/blob/main/google/cloud/bigtable/data/_sync_autogen/mutations_batcher.py#L151) and [MutationBatcherAsync](https://github.com/googleapis/python-bigtable/blob/main/google/cloud/bigtable/data/_async/mutations_batcher.py#L182)'s `table` argument was renamed to `target`**, since it also supports [Authorized View](https://github.com/googleapis/python-bigtable/pull/1034) instances. This matches the naming used in other classes (PR: https://github.com/googleapis/python-bigtable/pull/1153)
```
gkevinzheng, line 18: nit: I know that word wrapping on text files exists in IDEs, but could you split some of these longer lines into smaller lines? It shouldn't affect the final output at all but will make it easier to do things like read the raw file and whatnot.

PR #1187, opened 2025-08-11
feat: add data model for client side metrics

daniel-sanche:
This PR revives https://github.com/googleapis/python-bigtable/pull/923, which was de-priotirized to work on the sync client. This PR brings it back, working with both async and sync. It also adds a grpc interceptor, as an improved way to capture metadata across both clients

---

## Design

The main architecture looks like this:

300137129-bebbb05a-20f0-45c2-9d38-e95a314edf64 drawio (1)

Most of the work is done by the ActiveOperationMetric class, which is instantiated with each rpc call, and updated through the lifecycle of the call. When the rpc is complete, it will call `on_operation_complete` and `on_attempt_complete` on the MetricsHandler, which can then log the completed data into OpenTelemetry (or theoretically, other locations if needed)

Note that there are separate classes for active vs completed metrics (`ActiveOperationMetric`, `ActiveAttemptMetric`, `CompletedOperationMetric`, `CompletedAttemptMetric`). This is so that we can keep fields mutable and optional while the request is ongoing, but pass down static immutable copies once the attempt is completed and no new data is coming


google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,451 @@
...
+
+
+LOGGER = logging.getLogger(__name__)
+
+# default values for zone and cluster data, if not captured
+DEFAULT_ZONE = "global"
+DEFAULT_CLUSTER_ID = "unspecified"
+
+# keys for parsing metadata blobs
+BIGTABLE_METADATA_KEY = "x-goog-ext-425905942-bin"
```
mutianf: nit: maybe a more descriptive name like BIGTABLE_LOCATION_METADATA_KEY?

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,451 @@
...
+SERVER_TIMING_METADATA_KEY = "server-timing"
+SERVER_TIMING_REGEX = re.compile(r".*gfet4t7;\s*dur=(\d+\.?\d*).*")
+
+INVALID_STATE_ERROR = "Invalid state for {}: {}"
+
+
+class OperationType(Enum):
+    """Enum for the type of operation being performed."""
+
+    READ_ROWS = "ReadRows"
```
mutianf: there should also be a READ_ROW so we know if it's a point read or a scan

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,451 @@
...
+    # the time taken by the backend, in nanoseconds. Taken from response header
+    gfe_latency_ns: int | None = None
+    # time waiting on user to process the response, in nanoseconds
+    # currently only relevant for ReadRows
+    application_blocking_time_ns: int = 0
+    # backoff time is added to application_blocking_time_ns
+    backoff_before_attempt_ns: int = 0
+    # time waiting on grpc channel, in nanoseconds
+    # TODO: capture grpc_throttling_time
+    grpc_throttling_time_ns: int = 0
```
mutianf: we realized that in java this metric also doesn't capture the time a request queued on the channel. So if it's hard to get it in python we can skip it.

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,451 @@
...
+          - status: The status of the attempt.
+        """
+        if self.state != OperationState.ACTIVE_ATTEMPT or self.active_attempt is None:
+            return self._handle_error(
+                INVALID_STATE_ERROR.format("end_attempt_with_status", self.state)
+            )
+        if isinstance(status, BaseException):
+            status = self._exc_to_status(status)
+        complete_attempt = CompletedAttemptMetric(
+            duration_ns=time.monotonic_ns() - self.active_attempt.start_time_ns,
```
mutianf: should we add a sanity check to make sure this is >= 0? (in case there's bug in the code, we don't want csm to break client code)

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -0,0 +1,451 @@
...
+            )
+        final_status = (
+            self._exc_to_status(status) if isinstance(status, BaseException) else status
+        )
+        if self.state == OperationState.ACTIVE_ATTEMPT:
+            self.end_attempt_with_status(final_status)
+        self.was_completed = True
+        finalized = CompletedOperationMetric(
+            op_type=self.op_type,
+            uuid=self.uuid,
```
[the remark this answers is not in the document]
daniel-sanche: Good catch. In an older version of this code, I used a map to track active operations, with the uuid as the look-up key. But I changed it to use `contextvars`, which made everything a lot cleaner. So I think this can be removed now

google/cloud/bigtable/data/_metrics/handlers/_base.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+    Base class for all metrics handlers. Metrics handlers will receive callbacks
+    when operations and attempts are completed, and can use this information to
+    update some external metrics system.
+    """
+
+    def __init__(self, **kwargs):
+        pass
+
+    def on_operation_complete(self, op: CompletedOperationMetric) -> None:
```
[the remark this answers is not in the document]
daniel-sanche: This represents the split between recording metric data, and then exporting it somewhere else

- `ActiveOperationMetric` is the class that tracks an ongoing operation. It's passed around through an rpc's lifecycle, and is finalized by calling `operation.end_with_status()`
- `MetricHandler` represents a destination for the metrics, like OTel. The implementation of this is [mostly in a separate PR](https://github.com/googleapis/python-bigtable/pull/1189). The `ActiveOperationMetric` will call `metric_handler.on_operation_complete()` when it's time to export a new metric

google/cloud/bigtable/data/_metrics/tracked_retry.py
```diff
@@ -0,0 +1,133 @@
...
+
+    Should be passed as on_error callback
+    """
+
+    def wrapper(exc: Exception) -> None:
+        try:
+            # record metadata from failed rpc
+            if isinstance(exc, GoogleAPICallError) and exc.errors:
+                rpc_error = exc.errors[-1]
+                metadata = list(rpc_error.trailing_metadata()) + list(
```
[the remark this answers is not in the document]
daniel-sanche: I don't think so. 

The metrics_interceptor and tracked_retry are two separate methods for capturing metadata in different contexts, but the source of truth is always `operation.add_response_metadata()`, which they both write to

google/cloud/bigtable/data/_metrics/data_model.py
daniel-sanche: sounds good, maybe I'll remove this then

google/cloud/bigtable/data/_metrics/data_model.py
daniel-sanche: sounds good, I'll change this

google/cloud/bigtable/data/_metrics/data_model.py
mutianf: Ok, I think in java otel if we do otel.record(negative_number) it'll throw some exception. I think we can default to 0 and log a warning

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -64,7 +64,23 @@ class OperationType(Enum):
 
 
 class OperationState(Enum):
-    """Enum for the state of the active operation."""
+    """Enum for the state of the active operation.
+
```
mutianf: in end_with_status, it seems like CREATED can transition to COMPLETED, which I think it's valid if the rpc timed out before attempt is started (but probably rare), I think we should update the comment here?

mutianf (dismissed):

google/cloud/bigtable/data/_metrics/data_model.py
```diff
@@ -376,7 +387,8 @@ def end_with_success(self):
         """
         Called to mark the end of the operation with a successful status.
 
-        Assumes operation is not already in COMPLETED state.
+        StartState: CREATED | ACTIVE_ATTEMPT | BETWEEN_ATTEMPTS
```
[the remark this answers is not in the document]
daniel-sanche: - I think it only goes to BETWEEN_ATTEMPTS when an attempt fails. So if a request is successful, it can go straight from ACTIVE_ATTEMPT to COMPLETE
- For CREATED->COMPLETE, I don't think we'd expect that to happen often, but it IIRC, it should be possible to run into an exception or something in the handwritten code before the actual rpc starts. Maybe we'd just want to drop the metric in that case, since there isn't much to record?

google/cloud/bigtable/data/_metrics/data_model.py
daniel-sanche: Good point, I added it to the diagram

mutianf (approved):

PR #1188, opened 2025-08-11
feat: added client side metric instrumentation to basic rpcs

daniel-sanche:
This PR builds off of https://github.com/googleapis/python-bigtable/pull/1187 to add instrumentation to basic data client rpcs (check_and_mutate, read_modify_write, sample_row_keys, mutate_row)

Metrics are not currently being exported anywhere, just collected and dropped. A future PR will add a GCP exporter to the system


tests/system/data/test_metrics_async.py
```diff
@@ -0,0 +1,943 @@
...
+        assert operation.op_type.value == "MutateRow"
+        assert operation.is_streaming is False
+        assert len(operation.completed_attempts) == num_retryable + 1
+        assert operation.cluster_id == expected_cluster
+        assert operation.zone == expected_zone
+        # validate attempts
+        for i in range(num_retryable):
+            attempt = handler.completed_attempts[i]
+            assert isinstance(attempt, CompletedAttemptMetric)
+            assert attempt.end_status.name == "ABORTED"
```
mutianf: Not related to the metrics, but ABORTED shouldn't be retried for mutate row?

tests/system/data/test_metrics_async.py
```diff
@@ -0,0 +1,943 @@
...
+            assert attempt.gfe_latency_ns is None
+        final_attempt = handler.completed_attempts[num_retryable]
+        assert isinstance(final_attempt, CompletedAttemptMetric)
+        assert final_attempt.end_status.name == "PERMISSION_DENIED"
+        assert final_attempt.gfe_latency_ns is None
```
mutianf: I think as long as the request gets to the server, gfe_latency_ns should not be none. So this is probably related to the test setup ?

tests/system/data/test_metrics_async.py
daniel-sanche: It doesn't look like it, and it's not listed as a retryable error [in the service config.](https://github.com/googleapis/googleapis/blob/4b64cb41fe090cb27e540af23ee1a4f73d9c3118/google/bigtable/v2/bigtable_grpc_service_config.json#L75) Is it in Java? It would be easy to add here

tests/system/data/test_metrics_async.py
daniel-sanche: It looks like this test is testing the case where we have multiple transient retryable errors, before encountering a terminal error. For the errors, I'm  using a custom error_injector class, to control which errors are triggered in which sequence. So this test wouldn't ever be reaching the real backend

It's been a while since I looked at this, but I remember having issues fully controlling the backend errors in the way I needed for some of these tests, which is why I used the error_injector. There are other tests in this file that send unauthorized requests to trigger a real backend failure though.

parthea:

daniel-sanche:
Re-opening to see if we can get this in before the migration. If not, we can close and re-start this in the monorepo

parthea:
As per offline discussion with @daniel-sanche, closing PR to prepare for the code migration to google-cloud-python

PR #1189, opened 2025-08-11
[DRAFT] feat: client side metrics handlers

daniel-sanche:
This PR builds off of https://github.com/googleapis/python-bigtable/pull/1187 to add handlers to the client-side metrics system, which can subscribe to the metrics stream, and export the results into different collection systems

TODO:
- review comments in https://github.com/googleapis/python-bigtable/pull/924

---

We add three handlers to the system:
- `GoogleCloudMetricsHandler`: sends metrics to a private OpenTelemetry meter, and then periodically exports them to GCP. Built on top of `OpenTelemetryMetricsHandler`
- `OpenTelemetryMetricsHandler`: sends metrics to the root MeterProvider, so the user can access the exported metrics for their own systems. This will be off by default, but can be added alongside `GoogleCloudMetricsHandler` if needed
- `_StdoutMetricsHandler`: can print metrics to stdout as they arrive. Mostly for debugging (we can remove this if you don't think it's useful)

### DIFF
--- a/tests/system/data/test_metrics_async.py
+++ b/tests/system/data/test_metrics_async.py
@@ -135,14 +135,6 @@ class _ErrorInjectorInterceptor(
 
 @CrossSync.convert_class(sync_name="TestMetrics")
 class TestMetricsAsync(SystemTestRunner):
-    @CrossSync.drop
-    @pytest.fixture(scope="session")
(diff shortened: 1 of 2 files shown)


parthea:

PR #1201, opened 2025-08-26
docs: Added UPGRADING.md with admin client migration

gkevinzheng:
The changes to `UPGRADING.md` here builds upon some of the changes to `UPGRADING.md` in #1169 with additions for admin client migration.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,0 +1,36 @@
+# 3.0.0 Migration Guide
+
+The v3.0.0 release of `google-cloud-bigtable` deprecates the previous `google.cloud.bigtable.Client` class in favor of distinct clients for the two API surfaces, supporting both sync and async calls:
+- Data API: 
+  - `google.cloud.bigtable.data.BigtableDataClient`
+  - `google.cloud.bigtable.data.BigtableDataClientAsync`
+- Admin API: 
+  - `google.cloud.bigtable.admin.BigtableInstanceAdminClient`
+  - `google.cloud.bigtable.admin.BigtableInstanceAdminClientAsync`
+  - `google.cloud.bigtable.admin.BigtableTableAdminClient`
+  - `google.cloud.bigtable.admin.BigtableTableAdminClientAsync`
+
+The deprecated client will remain available as an alternative API surface, which internally delegates calls to the respective new clients. For most users, existing code will continue to work as before. But there may be some breaking changes associated with this update, which are detailed in this document.
+
+### Admin Client Migration
+
+While we do recommend you instantiate the new admin API clients directly, you can access the new admin API 
+clients via the `google.cloud.bigtable.Client` class:
+
+```
+from google.cloud.bigtable import Client
+
+client = Client()
+
+
+# google.cloud.bigtable.admin.BigtableInstanceAdminClient
+instance_admin_client = client.instance_admin_client()
+
+# google.cloud.bigtable.admin.BigtableTableAdminClient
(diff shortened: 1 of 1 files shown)


UPGRADING.md
```diff
@@ -0,0 +1,36 @@
...
+  - `google.cloud.bigtable.admin.BigtableInstanceAdminClient`
+  - `google.cloud.bigtable.admin.BigtableInstanceAdminClientAsync`
+  - `google.cloud.bigtable.admin.BigtableTableAdminClient`
+  - `google.cloud.bigtable.admin.BigtableTableAdminClientAsync`
+
+The deprecated client will remain available as an alternative API surface, which internally delegates calls to the respective new clients. For most users, existing code will continue to work as before. But there may be some breaking changes associated with this update, which are detailed in this document.
+
+### Admin Client Migration
+
+While we do recommend you instantiate the new admin API clients directly, you can access the new admin API
```
daniel-sanche, line 17: We should show how to do it the recommended way first. Maybe something like:

> The TableAdminClient can now be imported directly from google.cloud.bigtable.admin. This is the new recommended approach:
> `from google.cloud.bigtable.admin import TableAdminClient`
> ...
> Access to the admin clients through `google.cloud.bigtable.Client` is also supported for backward compatibility:
> ...

UPGRADING.md
```diff
@@ -0,0 +1,50 @@
...
+from google.cloud.bigtable import Client
+
+client = Client()
+
+# google.cloud.bigtable.admin.BigtableInstanceAdminClient
+instance_admin_client = client.instance_admin_client()
+
+# google.cloud.bigtable.admin.BigtableTableAdminClient
+table_admin_client = client.table_admin_client()
+```
```
iam-ethan-huang, line 45: What if user add new client with the future-proof import, in addition to existing client like the following?

i.e.

```
# User's existing code

from google.cloud.bigtable import Client
client = Client()
instance_admin_client = client.instance_admin_client()

# Added new code to try out the "correct way" of importing new client"
from google.cloud.bigtable.admin import BigtableInstanceAdminClient
instance_admin_client_new = BigtableInstanceAdminClient()
```

I guess this technically would create two clients which we don't want?

UPGRADING.md
```diff
@@ -0,0 +1,50 @@
...
+  - `google.cloud.bigtable.data.BigtableDataClientAsync`
+- Admin API: 
+  - `google.cloud.bigtable.admin.BigtableInstanceAdminClient`
+  - `google.cloud.bigtable.admin.BigtableInstanceAdminAsyncClient`
+  - `google.cloud.bigtable.admin.BigtableTableAdminClient`
+  - `google.cloud.bigtable.admin.BigtableTableAdminAsyncClient`
+
+The deprecated client will remain available as an alternative API surface, which internally delegates calls to the respective new clients. For most users, existing code will continue to work as before. But there may be some breaking changes associated with this update, which are detailed in this document.
+
+### Admin Client Migration
```
iam-ethan-huang, line 15: Currently it is more like a quick-start guide, instead of the migration guide.

For migration, we should start from what user have from legacy client and suggest what to do from there.

For example:
Path A (if you have time) - Add new import, remove old import, and convert all methods to new style, rebuild, test.
Path B (if you don't have time, but want to get new features) - Get new client via legacy client and use it. 

And indicate the path B is only for backward compatible phase and will be eventually deprecated.

iam-ethan-huang (changes requested):

PR #1204, opened 2025-09-04
chore: Moved admin client to google.cloud.bigtable.admin

### LINKED ISSUE #1155: Move admin client alongside data client
We should move the admin client code to live in `google.cloud.bigtable.admin`, along-side the data client. We wills till need to provide aliases so it can be used at the old path, but we also expect this to be a breaking change

gkevinzheng:
Changes:

1. Moved the admin library from `google.cloud.bigtable_admin/bigtable_admin_v2` to `google.cloud.bigtable.admin`. The autogenerated portion of the library was generated as follows:
```
py_gapic_library(
    name = "bigtable_admin_py_gapic",
    srcs = [":admin_proto"],
    grpc_service_config = "bigtableadmin_grpc_service_config.json",
    opt_args = [
        "python-gapic-namespace=google.cloud.bigtable",
        "python-gapic-name=admin",
    ],
    rest_numeric_enums = True,
    service_yaml = "bigtableadmin_v2.yaml",
    transport = "grpc+rest",
    deps = [
        "//google/iam/v1:iam_policy_py_proto",
    ],
)
```
This change will be committed close to when we will release the v3 client.

2. Created import aliasing to make sure that old imports from `bigtable_admin/bigtable_admin_v2` still work as intended.
1. Created unit tests for the import aliasing. Since there's an existing layer of import aliasing for objects within the GAPIC library, we create unit tests to make sure the composition of existing and new import aliasing works as intended.
1. Made the classic client use the new selective GAPIC `BigtableTableAdminClient` instead of the `BaseBigtableTableAdminClient` as its table admin client.

Fixes #1155


docs/index.rst
```diff
@@ -9,6 +9,7 @@ Client Types
 
    data_client/data_client_usage
    classic_client/usage
+   admin_client/admin_client_usage
```
daniel-sanche, line 12: I probably would have preferred to have the new docs as a separate PR to keep things focused, but probably not a big deal

google/cloud/bigtable/admin_v2/__init__.py
```diff
@@ -0,0 +1,274 @@
+# -*- coding: utf-8 -*-
```
daniel-sanche, line 1: do we have to have both `google.cloud.bigtable.admin` and `google.cloud.bigtable.admin_v2`? I thought we'd just need `google.cloud.bigtable.admin`

It seems confusing to have `google.cloud.bigtable.admin`, `google.cloud.bigtable.admin_v2`, `google.cloud.bigtable_admin`, and `google.cloud.bigtable.admin_v2`. Maybe we should reconsider

google/cloud/bigtable/admin_v2/overlay/services/bigtable_table_admin/__init__.py
```diff
@@ -1,22 +1,23 @@
...
 #
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-#
-from .client import BigtableTableAdminClient
+
+# TODO: Add the async client after owlbot changes.
```
daniel-sanche, line 15: What does this mean?

owlbot.py
```diff
@@ -158,4 +169,152 @@ def get_staging_dirs(
     """# todo(kolea2): temporary workaround to install pinned dep version
 INSTALL_LIBRARY_FROM_SOURCE = False""")
 
+# --------------------------------------------------------------------------
+# Admin Overlay work
+# --------------------------------------------------------------------------
```
daniel-sanche, line 174: I'm a bit confused why these new owlbot changes have to be added? Why are they only needed after this move?

samples/generated_samples/bigtableadmin_v2_generated_bigtable_instance_admin_create_app_profile_sync.py
```diff
@@ -0,0 +1,57 @@
+# -*- coding: utf-8 -*-
```
daniel-sanche, line 1: do we need to merge main into `v3_staging`?

daniel-sanche (changes requested):

daniel-sanche:
Maybe we should make a small doc to make sure we're all aligned on what the final structure will look like? I'd hate to spend a lot of time iterating here if we have to move things back and forth a bunch

Good to see import aliases aren't going to be a problem though

google/cloud/bigtable/admin_v2/__init__.py
gkevinzheng: The GAPIC structure is that `google.cloud.bigtable.admin` aliases to `google.cloud.bigtable.admin_v<version>`

owlbot.py
gkevinzheng: `v3_staging` doesn't have the selective GAPIC client in it right now, so I added it back in.

daniel-sanche:
It looks like the overlay and utils are duplicated in `google/cloud/bigtable/admin` and `google/cloud/bigtable_admin_v2`. Is that intentional?

google/cloud/bigtable_admin/__init__.py
```diff
@@ -13,439 +13,12 @@
...
-    "Table",
-    "RestoreSourceType",
-    "Type",
-)
-
-import google.cloud.bigtable_admin_v2.overlay  # noqa: F401
-from google.cloud.bigtable_admin_v2.overlay import *  # noqa: F401, F403
-
-__all__ += google.cloud.bigtable_admin_v2.overlay.__all__
+__all__ = google.cloud.bigtable.admin.__all__
```
daniel-sanche: Do we have test that try importing in different ways, to make sure they all work the same as before?

i.e. `import google.cloud.bigtable.admin.Type`, `from google.cloud.bigtable.admin import Type`, etc

owlbot.py
```diff
@@ -86,15 +85,15 @@ def get_staging_dirs(
     s.move(library / "scripts")
 
 for library in get_staging_dirs(bigtable_admin_default_version, "bigtable_admin"):
-    is_fresh_admin_copy = \
-        s.move(library / "google/cloud/bigtable_admin", excludes=["**/gapic_version.py"])
     is_fresh_admin_v2_copy = \
-        s.move(library / "google/cloud/bigtable_admin_v2", excludes=["**/gapic_version.py"])
-    s.move(library / "tests")
+        s.move(library / "google/cloud/bigtable/admin_v2",
```
parthea: Do we need a CL to change `google/cloud/bigtable_admin_v2` to `google/cloud/bigtable/admin_v2`? 

Currently we generate `google/cloud/bigtable_admin_v2`

https://github.com/googleapis/googleapis-gen/tree/master/google/bigtable/admin/v2/bigtable-admin-v2-py/google/cloud/bigtable_admin_v2

If we want `google/cloud/bigtable/admin_v2` then we'll need to delete these lines
https://github.com/googleapis/googleapis/blob/a17b84add8318f780fcc8a027815d5fee644b9f7/google/bigtable/admin/v2/BUILD.bazel#L173-L174

google/cloud/bigtable_admin_v2/__init__.py
```diff
@@ -13,262 +13,24 @@
...
-
-__all__ += overlay.__all__  # noqa: F405
+# iterate and import all submodules to populate sys.modules
+for _, name, _ in pkgutil.walk_packages(
+    path=google.cloud.bigtable.admin.__path__,
+    prefix=_NEW_PATH + ".",
+):
+    mod = importlib.import_module(name)
+    alias = name.replace(_NEW_PATH, __name__, 1)
+    sys.modules[alias] = mod
```
daniel-sanche: I'm curious why this is so different than `google/cloud/bigtable_admin/__init__.py`. Aren't they both aliases?

daniel-sanche:
You should also consider moving the classic client alongside the others, something like this: https://github.com/googleapis/python-bigtable/pull/1200

Probably not in this PR though. We can consider that near the end

daniel-sanche (approved):
LGTM

PR #1206, opened 2025-09-09
feat: add basic interceptor to client

daniel-sanche:
This PR adds a new grpc interceptor to the client, to be used for metrics collection. Currently, the interceptor is a no-op. Interception logic will be added in a future change.


chalmerlowe (dismissed):
I am approving, pending:

* See the #QUESTIONs and determine whether any changes need to be made, if no, then carry on.
* See the #PREFERENCEs and #NITs and see if there is merit to including them. If no, then carry on.

chalmerlowe (approved):
LGTM

PR #1216, opened 2025-10-03
fix: Fix instance registration cleanup on early iterator termination

jackdingilian:
This makes two related changes to attempt to address https://github.com/googleapis/google-cloud-python/issues/15283

1) Wrap the generator used for execute query in try-finally and call close in the finally block. This cleans up the instance registration used by ping and warm even when iteration fails or is exited early.

2) Change the instance registration method so that it does not need to take a reference to the owner. When an iterator is abandoned early, with no error, finally won't run until it is GC'ed. This reference in the registration future prevents GC though. Changing this to take the id and app profile directly resolves the issue.


google/cloud/bigtable/data/_async/client.py
```diff
@@ -486,13 +487,15 @@ async def _register_instance(
 
         Args:
           instance_id: id of the instance to register.
-          owner: table that owns the instance. Owners will be tracked in
+          app_profile_id: id of the app profile calling the instance.
+          owner_id: integer id of the object owning the instance. Owners will be tracked in
```
daniel-sanche, line 491: nit: maybe reference that this uses the id() identity function?

Doesn't matter too much since this is internal, but it might not be clear what id means in this context. Especially since we have table_id, app_profile_id, instance_id, etc

daniel-sanche (dismissed):
I left a small docstring comment, but LGTM

google/cloud/bigtable/data/_async/client.py
jackdingilian: Updated

daniel-sanche (approved):

PR #1218, opened 2025-10-14
tests: System/compatibility tests for testing shim compatiblity

### LINKED ISSUE #1156: Add new system tests
There is a severe lack of end-to-end system tests for the legacy client (described in [go/bigtable_legacy_deprecation](http://goto.google.com/bigtable_legacy_deprecation)). Before implementing the shim, we should dedicate some time to make sure we have appropriate coverage of the existing features

gkevinzheng:
Additional system tests for testing shim compatibility with the classic client interface.

Also fixes a bug where encountering a retriable error after another retriable error mid-stream raised an exception instead of retrying.

Fixes #1156


google/cloud/bigtable/row_data.py
```diff
@@ -227,7 +227,7 @@ def _on_error(self, exc):
             retry_request = self._create_retry_request()
 
         self._row_merger = _RowMerger(self._row_merger.last_seen_row_key)
-        self.response_iterator = self.read_method(retry_request)
+        self.response_iterator = self.read_method(retry_request, retry=self.retry)
```
daniel-sanche: why was this change needed?

tests/system/v2_client/test_data_api.py
```diff
@@ -26,9 +34,101 @@
...
+PASS_ALL_FILTER = row_filters.PassAllFilter(True)
+BLOCK_ALL_FILTER = row_filters.BlockAllFilter(True)
+
+READ_ROWS_METHOD_NAME = "/google.bigtable.v2.Bigtable/ReadRows"
+
+
+class SelectiveMethodsErrorInjector(UnaryStreamClientInterceptor):
+    def __init__(self):
+        # As long as there are more items on this list, the items on the list
+        # are as follows:
```
daniel-sanche: I assume after the list is exhausted, it always behaves as normal?

tests/system/v2_client/test_data_api.py
```diff
@@ -59,6 +159,38 @@ def rows_to_delete():
...
+@pytest.fixture(scope="function")
+def data_table_read_rows_retry_tests(data_table, rows_to_delete):
+    row_keys = [f"row_key_{i}".encode() for i in range(0, 32)]
+    columns = [f"col_{i}".encode() for i in range(0, 32)]
+
+    _populate_table(
+        data_table, rows_to_delete, row_keys, columns, CELL_VAL_READ_ROWS_RETRY
+    )
+
+    yield data_table
```
daniel-sanche: Isn't this re-populating the table before each test function? Is that necessary?

daniel-sanche (commented):
A couple comments for now, but I'll look at the tests in more detail soon

tests/system/v2_client/test_data_api.py
gkevinzheng: Let me refactor the `read_rows` system tests to reconfigure the error injector instead of re-populating the table before each test function.

google/cloud/bigtable/row_data.py
gkevinzheng: This change was needed because the `read_rows` method that gets passed in gets called with the GAPIC default retry instead of the retry that the user passed in or `DEFAULT_READ_RETRY` for this instance. If there's an error with this specific instance of `self.read_method` it doesn't get retried and errors out.

tests/system/v2_client/test_data_api.py
```diff
@@ -59,6 +159,27 @@ def rows_to_delete():
...
+
+    for row in rows_to_delete:
+        row.clear()
+        row.delete()
+        row.commit()
```
daniel-sanche: please use a try...finally block to make sure rows are fully deleted. And do the same with the table created in data_table

daniel-sanche (approved):
LGTM

PR #1238, opened 2025-12-03
feat: initializing the new data client

gkevinzheng:
Changes:
- Added a new property `new_table_data_client` (might need to be renamed) in the classic client for the new table data client
- Added two hidden kwargs (`client_info` and `disable_background_channel_refresh`) to the new data client constructor to support the classic client.

Fixes googleapis/google-cloud-python#15298


google/cloud/bigtable/client.py
```diff
@@ -369,6 +374,23 @@ def instance_admin_client(self):
             self._instance_admin_client = klass(self)
         return self._instance_admin_client
 
+    @property
+    def new_table_data_client(self):
+        """Getter for the new Data Table API.
+
+        TODO: Replace table_data_client with this implementation
+        when shimming legacy client is finished.
```
daniel-sanche: I don't think that's a goal. We should try to keep returning what we did before

google/cloud/bigtable/client.py
```diff
@@ -146,6 +150,7 @@ class Client(ClientWithProject):
     _table_data_client = None
     _table_admin_client = None
     _instance_admin_client = None
+    _new_table_data_client = None
```
daniel-sanche: 1. We should avoid calling this "new", because someday it will be old
2. We don't need internal references for both the gapic client and the veneer client, since the veneer holds a copy of the gapic. 

So we can probably remove _table_data_client and _new_table_data_client, and just have a `_data_client`, that holds the veneer.

Then in `def table_data_client(self):`, return `self._data_client._gapic_client`

google/cloud/bigtable/data/_async/client.py
```diff
@@ -236,6 +236,9 @@ def __init__(
                 "is the default."
             )
         self._is_closed = CrossSync.Event()
+        self._disable_background_channel_refresh = bool(
```
daniel-sanche: Add a comment here, explaining that this is a private argument for internal use only

google/cloud/bigtable/client.py
```diff
@@ -54,6 +54,11 @@
 from google.cloud.bigtable.cluster import _CLUSTER_NAME_RE
```
daniel-sanche: In https://github.com/googleapis/python-bigtable/issues/1157, I also suggested adding a `client.table()` reference as part of the [Table class](https://github.com/googleapis/python-bigtable/blob/044efe74e0ae8eae46d5567276b5a5d9b699c315/google/cloud/bigtable/table.py#L129). You should probably do that in this PR as well

tests/unit/data/_async/test_client.py
```diff
@@ -245,6 +309,19 @@ async def test__start_background_channel_refresh(self):
...
             await client.close()
 
+    @CrossSync.pytest
+    async def test__start_background_channel_refresh_disable_refresh(self):
+        client = self._make_client(
+            project="project-id",
+            disable_background_channel_refresh=True,
+        )
+        # should create background tasks for each channel
+        with mock.patch.object(client, "_ping_and_warm_instances", CrossSync.Mock()):
```
daniel-sanche: Do you need to mock this? I think that was done in other tests as a crude way of disabling background refresh

google/cloud/bigtable/table.py
```diff
@@ -132,6 +132,15 @@ def __init__(self, table_id, instance, mutation_timeout=None, app_profile_id=Non
         self._app_profile_id = app_profile_id
         self.mutation_timeout = mutation_timeout
 
+        self._table_impl = (
+            self._instance._client._data_client.get_table(
+                self._instance.instance_id,
+                self.table_id,
+            )
+            if self._instance
+            else None
```
daniel-sanche: Why do you need `if self._instance`? It doesn't seem optional?

I don't think we'd ever want this to be None, right?

daniel-sanche (commented):
I added a couple suggestions, but let me know if I'm missing context

daniel-sanche (approved):

PR #1257, opened 2026-01-16
feat: Rerouted CheckAndMutateRows and ReadModifyWriteRows

gkevinzheng:
**Changes made:**

- `Row` objects hold `Mutation` and `ReadModifyWriteRowRule` objects from the data client rather than raw protos.
- Rerouted `ConditionalRow.commit` and `AppendRow.commit` (CheckAndMutateRows and ReadModifyWriteRows respectively) to use the data client, or more specifically, `self._table._table_impl`
- Added function `DirectRow._to_mutation_pbs` for retrieving mutations in proto form for the current `MutateRows` implementation, as well as for `DirectRow.get_mutations_size`.
- Removed unnecessary helper functions and tests for helper functions

### DIFF
--- a/google/cloud/bigtable/row.py
+++ b/google/cloud/bigtable/row.py
@@ -14,16 +14,13 @@
 
 """User-friendly container for Google Cloud Bigtable Row."""
 
-
-import struct
-
 from google.cloud._helpers import _datetime_from_microseconds  # type: ignore
 from google.cloud._helpers import _microseconds_from_datetime  # type: ignore
 from google.cloud._helpers import _to_bytes  # type: ignore
-from google.cloud.bigtable_v2.types import data as data_v2_pb2
 
+from google.cloud.bigtable.data import mutations
+from google.cloud.bigtable.data import read_modify_write_rules as rmw_rules
 
-_PACK_I64 = struct.Struct(">q").pack
 
 MAX_MUTATIONS = 100000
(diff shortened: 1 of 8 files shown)


google/cloud/bigtable/row.py
```diff
@@ -222,27 +212,28 @@ def _delete_cells(self, column_family_id, columns, time_range=None, state=None):
...
-            mutations_list.append(mutation_pb)
         else:
-            delete_kwargs = {}
+            start_timestamp_micros = None
+            end_timestamp_micros = None
             if time_range is not None:
-                delete_kwargs["time_range"] = time_range._to_pb()
+                timestamps = time_range._to_dict()
+                start_timestamp_micros = timestamps.get("start_timestamp_micros")
+                end_timestamp_micros = timestamps.get("end_timestamp_micros")
```
daniel-sanche: nit: this logic might be a bit cleaner like this:
```
timerange_dict = time_range._to_dict() if time_range else {}
start_timestamp_micros = timestamps.get("start_timestamp_micros")
end_timestamp_micros = timestamps.get("end_timestamp_micros")
```

But it's fine as-is too

daniel-sanche (commented):
A couple small comments. The only important one is dealing with `mutations_list`

This is looking really good though, it's great seeing all this code removed!

daniel-sanche (approved):

daniel-sanche (approved):

PR #1287, opened 2026-02-02
[DRAFT] feat: use RetryInfo metadata from retryable exceptions

daniel-sanche:
Proof of concept of the RetryInfo feature. Allows the backend to set retry backoff settings

TODO: 
- the python backoff method [uses truncated backoff](https://github.com/googleapis/python-api-core/blob/a6a5b6fe788610a9fe00a276615f3f7d3d8be7ce/google/api_core/retry/retry_base.py#L100), but the conformance tests enforce a minimum value. This is causing one of the conformance tests to be flaky
- Do we only need the first retry info? Or do we expect multiple?
- Feature flags?
- implement disable feature
- tests


parthea:

PR #1290, opened 2026-02-05
feat: Reworked MutateRows to use the data client

gkevinzheng:
**Changes Made:**

- Rerouted MutateRows to use the data client's `Table.bulk_mutate_rows` function.
- Reworked error handling in `Table.mutate_rows`. Since we must return a list of `Status` objects, we process the list of errors obtained from the `_MutateRowsOperation` back into a corresponding list of error statuses.
- Reworked retries in `Table.mutate_rows` as follows:
  - `retry.deadline` -> `operation_timeout`
  - `timeout`/`mutation_timeout` -> `attempt_timeout`
  -  A null retry deadline now defaults to the table default mutate rows operation timeout instead of being defined as no timeout. The docstring has been updated to reflect that. A retry deadline of 0.0 will still be a no-retry option, and is implemented by setting the list of retriable errors to the empty list, and the operation timeout to the default mutate rows operation timeout.
- A `timestamp` of `None` in `set_cell`, which is the default, will now generate a client-side timestamp representing the current time, rather than defaulting to using the server-side timestamp. The docstrings for this have been updated to reflect this.

### DIFF
--- a/google/cloud/bigtable/table.py
+++ b/google/cloud/bigtable/table.py
@@ -31,7 +31,10 @@ from google.cloud.bigtable.backup import Backup
 from google.cloud.bigtable.column_family import _gc_rule_from_pb
 from google.cloud.bigtable.column_family import ColumnFamily
 from google.cloud.bigtable.data._helpers import TABLE_DEFAULT
-from google.cloud.bigtable.data.exceptions import MutationsExceptionGroup
+from google.cloud.bigtable.data.exceptions import (
+    RetryExceptionGroup,
+    MutationsExceptionGroup,
+)
 from google.cloud.bigtable.data.mutations import RowMutationEntry
 from google.cloud.bigtable.batcher import MutationsBatcher
 from google.cloud.bigtable.batcher import FLUSH_COUNT, MAX_MUTATION_SIZE
@@ -740,18 +743,20 @@ class Table(object):
         if timeout is DEFAULT:
             timeout = self.mutation_timeout
 
-        # To adhere to the retry strategy of do-nothing being achievable with a deadline
-        # of 0.0, we modify the retryable errors to be empty if such a deadline is passed.
         retryable_errors = RETRYABLE_MUTATION_ERRORS
-        operation_timeout = retry.deadline
 
         # The data client cannot take in zero or null values for deadline, so we set it to
-        # the default if that is the case. It shouldn't affect the behavior of the retry
-        # if a 0.0 deadline is set.
-        if not retry.deadline:
+        # the default if that is the case.
+        if retry.deadline is None:
             operation_timeout = TABLE_DEFAULT.MUTATE_ROWS
-            if retry.deadline == 0.0:
-                retryable_errors = []
+
+        # To adhere to the retry strategy of do-nothing being achievable with a deadline
+        # of 0.0, we modify the retryable errors to be empty if such a deadline is passed.
+        elif retry.deadline == 0:
+            operation_timeout = TABLE_DEFAULT.MUTATE_ROWS
+            retryable_errors = []
+        else:
+            operation_timeout = retry.deadline
 
         attempt_timeout = timeout
         mutation_entries = [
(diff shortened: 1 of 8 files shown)


google/cloud/bigtable/table.py
```diff
@@ -731,18 +734,66 @@ def mutate_rows(self, rows, retry=DEFAULT_RETRY, timeout=DEFAULT):
...
+        retryable_errors = RETRYABLE_MUTATION_ERRORS
+        operation_timeout = retry.deadline
+
+        # The data client cannot take in zero or null values for deadline, so we set it to
+        # the default if that is the case. It shouldn't affect the behavior of the retry
+        # if a 0.0 deadline is set.
+        if not retry.deadline:
+            operation_timeout = TABLE_DEFAULT.MUTATE_ROWS
+            if retry.deadline == 0.0:
+                retryable_errors = []
```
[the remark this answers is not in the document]
daniel-sanche: I agree this looks a bit cleaner

I'd also suggest doing `retry.deadline == 0`, instead of the float comparison

google/cloud/bigtable/table.py
```diff
@@ -731,18 +734,66 @@ def mutate_rows(self, rows, retry=DEFAULT_RETRY, timeout=DEFAULT):
...
-        return retryable_mutate_rows(retry=retry)
+
+        try:
+            operation.start()
+        except MutationsExceptionGroup:
+            # Take the first exception for each error index with a gRPC status code
+            # and set the status of that row entry to that grpc status. if none of the
+            # errors for a given index have gRPC status codes, return an UNKNOWN status
+            # with the first error message of each index.
+            for idx, errors in operation.errors.items():
```
daniel-sanche: You shouldn't have to go into the operation object like this. Can't you do something like:

```
try:
    self._table_impl.mutate_rows(
            mutation_entries,
            operation_timeout=operation_timeout,
            attempt_timeout=attempt_timeout,
            retryable_errors=retryable_errors,
        )
except MutationsExceptionGroup as exc_group:
    for error: FailedMutationEntryError in exc_group.exceptions:
        return_statuses[error.index] = status_pb2.Status(
            code=error.__cause__.grpc_status_code.value[0],
            message=error.__cause__.message,
            details=error.__cause__.details,
        )
return_statuses
```

(It looks like it might be a bit more complicated, since MutationExceptionGroup [doesn't guarantee all exceptions](https://github.com/googleapis/python-bigtable/blob/1c39a7de6ea161258d68a09f19eb1c3e5bd85ac9/google/cloud/bigtable/data/exceptions.py#L148) are FailedMutationEntryErrors. But looking at the [only place this exception is created](https://github.com/googleapis/python-bigtable/blob/1c39a7de6ea161258d68a09f19eb1c3e5bd85ac9/google/cloud/bigtable/data/_sync_autogen/_mutate_rows.py#L120), maybe that comment is wrong and should be changed?)

daniel-sanche (changes requested):

google/cloud/bigtable/table.py
```diff
@@ -731,18 +737,78 @@ def mutate_rows(self, rows, retry=DEFAULT_RETRY, timeout=DEFAULT):
...
+            #
+            # 2. In the case of a singular exception, if the error does not have a gRPC status
+            #    code, we return a status code of UNKNOWN.
+            #
+            # 3. In the case of a RetryExceptionGroup, we use terminal exception in the exception
+            #    group and process that.
+            for error in mut_exc_group.exceptions:
+                cause = error.__cause__
+                if isinstance(cause, RetryExceptionGroup):
+                    return_statuses[error.index] = self._get_status(
```
daniel-sanche, line 791: This `.index` assumes that all exceptions are type `FailedMutationEntryError`. Which  I think is right, but technically not what the typing says (see my comment [here](https://github.com/googleapis/python-bigtable/pull/1290#discussion_r2776525011))

Maybe if you do this, you should change `MutationsExceptionGroup` to promise it'll only hold `FailedMutationEntryError`? Otherwise this could break if the logic changes in the future

Alternatively, you could change this code to handle other error types. But I kind of prefer the first option

daniel-sanche (approved):

PR #1298, opened 2026-02-24
feat: Added rst_stream exception handling for ReadRows.

gkevinzheng:
According to go/rst_stream, `INTERNAL` errors with error messages related to an `rst_stream` error should be interpreted as `UNAVAILABLE` errors instead of internal errors. This PR creates a custom retry predicate to allow retrying of `INTERNAL` errors with rst_stream specific error messages if the `ServiceUnavailable` exception is allowed to be retried.

### DIFF
--- a/google/cloud/bigtable/data/_helpers.py
+++ b/google/cloud/bigtable/data/_helpers.py
@@ -132,31 +132,31 @@ def _retry_exception_factory(
     return source_exc, cause_exc
 
 
-def _read_rows_predicate_with_exceptions(*exception_types: type[Exception]) -> Callable[[Exception], bool]:
-    """A custom retry predicate for ReadRows.
-    
+def _rst_stream_aware_predicate(
+    *exception_types: type[Exception],
+) -> Callable[[Exception], bool]:
+    """A custom retry predicate.
+
     This predicate treats Internal error messages with RST_STREAM errors as
     ServiceUnavailable errors and will retry them if the Unavailable exception is retryable.
 
     Args:
         exception_types: Exception types to be retried during operation
-    
+
     Returns:
         Callable[[Exception], bool]: A retry predicate that takes in an exception and
             returns whether or not that exception is retryable
     """
-    is_exception_type = retries.if_exception_type(*exception_types)
-    
-    def predicate(exception: Exception) -> bool:
-        return (isinstance(exception, core_exceptions.InternalServerError) and any(m in exception.message.lower() for m in _RETRYABLE_INTERNAL_ERROR_MESSAGES)) or is_exception_type(exception)
+    # predicate to check for retryable error types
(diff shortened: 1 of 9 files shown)


google/cloud/bigtable/data/_helpers.py
```diff
@@ -122,6 +132,33 @@ def _retry_exception_factory(
     return source_exc, cause_exc
 
 
+def _read_rows_predicate_with_exceptions(*exception_types: type[Exception]) -> Callable[[Exception], bool]:
```
daniel-sanche: can we call this something more specific, since it's not really tied to read_rows? Maybe _rst_stream_aware_predicate?

google/cloud/bigtable/data/_helpers.py
```diff
@@ -122,6 +132,33 @@ def _retry_exception_factory(
...
+    def predicate(exception: Exception) -> bool:
+        return (isinstance(exception, core_exceptions.InternalServerError) and any(m in exception.message.lower() for m in _RETRYABLE_INTERNAL_ERROR_MESSAGES)) or is_exception_type(exception)
+
+    # Treating RST_STREAM internal errors as unavailable errors is only done if ServiceUnavailable is one of the
+    # given exception types. If InternalServerError is also a retryable exception, we don't necessarily need the
+    # custom predicate either.
+    if core_exceptions.ServiceUnavailable in exception_types and core_exceptions.InternalServerError not in exception_types:
+        return predicate
+
+    return is_exception_type
```
daniel-sanche: I find the logic a bit hard to follow, with the multiple predicate functions being passed around. Can we do something like this, to put all the cases in one place?

```
def _read_rows_predicate_with_exceptions(*exception_types: type[Exception]) -> Callable[[Exception], bool]:
    def predicate(exception: Exception)  -> bool:
        if isinstance(exception, exception_types):
           # handle standard retryable errors
           return True
        elif core_exceptions.ServiceUnavailable in exception_types and isinstance(exception, InternalServerError) and any(m in exception.message.lower() for m in _RETRYABLE_INTERNAL_ERROR_MESSAGES)):
            # special case: treat InternalServerError with specific message as ServiceUnavailable
            return True
        else:
            return False
    return predicate
```

or even this?

```
def _read_rows_predicate_with_exceptions(*exception_types: type[Exception]) -> Callable[[Exception], bool]:
    # predicate to check for retryable error types
    if_exception_type = retries.if_exception_type(*exception_types)
    # special case: treat InternalServerError with specific message as ServiceUnavailable
    rst_check = lambda e: core_exceptions.ServiceUnavailable in exception_types and isinstance(e, InternalServerError) and any(m in e.message.lower() for m in _RETRYABLE_INTERNAL_ERROR_MESSAGES))
  
   return lambda e: if_exception_type(e) or rst_check(e)
```

daniel-sanche (approved):

PR #1299, opened 2026-02-26
feat: Rerouted ReadRows to data client

gkevinzheng:
**Changes Made:**
- Added methods to convert `Row` and `Cell` objects in the data client to `PartialRowData` and `Cell` objects in the legacy client.
- Removed legacy client code related to processing `ReadRowResponse` chunks and testing `ReadRowResponse` chunks.
- Removed `_update_message_request` from `RowSet` because it's no longer needed to create a `ReadRowQuery`
- Rerouted `read_row` and `read_rows` to use their data client counterparts in `table.py`.


google/cloud/bigtable/row_data.py
```diff
@@ -108,94 +71,20 @@ class PartialRowsData(object):
...
-
-    :type retry: :class:`~google.api_core.retry.Retry`
-    :param retry: (Optional) Retry delay and deadline arguments. To override,
-                  the default value :attr:`DEFAULT_RETRY_READ_ROWS` can be
-                  used and modified with the
-                  :meth:`~google.api_core.retry.Retry.with_delay` method
-                  or the
-                  :meth:`~google.api_core.retry.Retry.with_deadline` method.
+    :type generator: :class:`Iterable[Row]`
+    :param generator: The `Row` iterator from :meth:`Table.read_rows`
```
daniel-sanche: You should add a note that this is not intended to be created directly

google/cloud/bigtable/row_data.py
```diff
@@ -108,94 +71,20 @@ class PartialRowsData(object):
...
-        self.response_iterator = read_method(
-            request, timeout=self.retry._deadline + 1, retry=self.retry
-        )
-
+    def __init__(self, generator):
+        self._generator = generator
         self.rows = {}
 
         # Flag to stop iteration, for any reason not related to self.retry()
         self._cancelled = False
```
daniel-sanche: is _cancelled still in use?

google/cloud/bigtable/row_data.py
```diff
@@ -108,94 +71,20 @@ class PartialRowsData(object):
...
-        else:
-            raise RuntimeError("unexpected internal state: " + self._)
-
     def cancel(self):
         """Cancels the iterator, closing the stream."""
-        self._cancelled = True
-        self.response_iterator.cancel()
+        self._generator.close()
 
     def consume_all(self, max_loops=None):
```
daniel-sanche, line 89: it looks like max_loops isnt used?

google/cloud/bigtable/row_data.py
```diff
@@ -212,169 +101,20 @@ class as a generator instead.
...
-
-    def _read_next_response(self):
-        """Helper for :meth:`__iter__`."""
-        resp_protoplus = self.retry(self._read_next, on_error=self._on_error)()
-        # unwrap the underlying protobuf, there is a significant amount of
-        # overhead that protoplus imposes for very little gain. The protos
-        # are not user visible, so we just use the raw protos for merging.
-        return data_messages_v2_pb2.ReadRowsResponse.pb(resp_protoplus)
-
     def __iter__(self):
```
daniel-sanche, line 104: What happens if you try to iterate over this multiple times? I assume since the inner generator is exhausted, it would just yield nothing? Is that similar to the old implementation?

Also, what do we expect to happen if we keep iterating after cancellation?

google/cloud/bigtable/row_data.py
gkevinzheng: Removed self._cancelled.

google/cloud/bigtable/row_data.py
gkevinzheng: I believe that in the original, if we keep iterating after cancellation, the `while not self._cancelled` part will return nothing or at most 1 item. I'll look into what happens when we iterate over the same `PartialRowsData` object multiple times.

google/cloud/bigtable/row_data.py
gkevinzheng: I'm not sure what to do here because it wasn't used in the original implementation of the function either, so I assume it was there to prevent a breaking change. Now that we are able to make a breaking change, I should remove it, right?

google/cloud/bigtable/row_data.py
gkevinzheng: As I previously suspected, iterating over the same `PartialRowsData` will yield nothing the second time around in the old implementation as well. The following system test passes when I test an older branch of `v3_staging`:

```

def test_table_read_rows_multiple_reads(
    data_table_read_rows_retry_tests,
):
    from types import SimpleNamespace

    rows_data = data_table_read_rows_retry_tests.read_rows()
    first_iteration = SimpleNamespace()
    first_iteration.rows = {}

    second_iteration = SimpleNamespace()
    second_iteration.rows = {}
    for item in rows_data:
        first_iteration.rows[item.row_key] = item
    
    for item in rows_data:
        second_iteration.rows[item.row_key] = item
    
    _assert_data_table_read_rows_retry_correct(first_iteration)
    assert second_iteration.rows == {}
```

I will add this to this branch to test that the same behavior occurs.

google/cloud/bigtable/row_data.py
```diff
@@ -108,94 +71,18 @@ class PartialRowsData(object):
...
-    :type retry: :class:`~google.api_core.retry.Retry`
-    :param retry: (Optional) Retry delay and deadline arguments. To override,
-                  the default value :attr:`DEFAULT_RETRY_READ_ROWS` can be
-                  used and modified with the
-                  :meth:`~google.api_core.retry.Retry.with_delay` method
-                  or the
-                  :meth:`~google.api_core.retry.Retry.with_deadline` method.
+    :type generator: :class:`Iterable[Row]`
+    :param generator: The `Row` iterator from :meth:`Table.read_rows`. This is not intended
+                      to be created directly.
```
daniel-sanche: Sorry, for the "This is not intended to be created directly." comment, I meant for the constructor itself, not just for the generator argument

Something like:

```
"""Convenience wrapper for consuming a ``ReadRows`` streaming response.

This class will be returned by the read_rows method, and should not be constructed manually

...
```

google/cloud/bigtable/row_data.py
daniel-sanche: Yeah, I think we can remove it

If you want to keep it around to be safe, that's fine. But try to give it a better docstring if you do

google/cloud/bigtable/row_data.py
daniel-sanche: Ok, that sounds fine then

daniel-sanche (commented):
Overall looks good, but you should either remove max_rows, or give it a better deprecation string

google/cloud/bigtable/row_data.py
```diff
@@ -92,7 +94,9 @@ def consume_all(self, max_loops=None):
            class as a generator instead.
 
         :type max_loops: int
-        :param max_loops: (Deprecated). This parameter does nothing.
+        :param max_loops: (Deprecated). Maximum number of times to try to consume
+                          an additional ``ReadRowsResponse``. This parameter is
+                          deprecated and is only kept for backwards compatibility.
```
daniel-sanche, line 99: nit: "backwards compatibility" kind of sounds like it still works. Maybe you should mention that it is not used?

daniel-sanche (approved):

PR #1308, opened 2026-03-06
feat: Added a batch completed callback to the data client mutations batcher

gkevinzheng:
**Changes made:**

- Refactored logic from `Table.mutate_rows` from producing a list of `Status` protos from a `MutationsExceptionGroup`
- Added private keyword argument for a batch completion callback in the MutationsBatcher.
- Added unit tests/system tests.


google/cloud/bigtable/data/_async/mutations_batcher.py
```diff
@@ -223,6 +230,7 @@ def __init__(
         batch_attempt_timeout: float | None | TABLE_DEFAULT = TABLE_DEFAULT.MUTATE_ROWS,
         batch_retryable_errors: Sequence[type[Exception]]
         | TABLE_DEFAULT = TABLE_DEFAULT.MUTATE_ROWS,
+        _batch_completed_callback: Optional[Callable[list[status_pb2.Status]]] = None,
```
daniel-sanche: I don't think we should add hidden arguments to the public constructor args like this. 
1. If we absolutely have to, we can add a **kwargs section, and pass it there
2. Ideally, we can have it set to None by default, and the shim can attach at right after init

google/cloud/bigtable/data/_async/mutations_batcher.py
```diff
@@ -380,6 +389,7 @@ async def _execute_mutate_rows(
                 list of FailedMutationEntryError objects for mutations that failed.
                 FailedMutationEntryError objects will not contain index information
         """
+        statuses = [status_pb2.Status(code=code_pb2.Code.OK)] * len(batch)
```
daniel-sanche: It feels risky to me to assume all statuses are successful before starting the operation. Can we set them to unknown at first, and then add an `else` case to the try/catch that sets them all to OK?

And then maybe we can change `_populate_statuses_from_mutations_exception_group` into `_create_statues_from_exception(batch, e)`, and it can return a full list indicating which ones succeeded as well as failed?

daniel-sanche (changes requested):

google/cloud/bigtable/data/_async/mutations_batcher.py
gkevinzheng: If we change `_populate_statuses_from_mutations_exception_group` to `_create_statuses_from_exception(batch, e)`, wouldn't the helper function also do the same thing of assuming the statuses are all OK until proven otherwise?

daniel-sanche: My concern is more about the sequence of events. If we wait until the operation is complete and we were given an error, it's safer to assume the others succeeded. But making the assumption before starting the operation, feels like it could lead to problems down the line.

To be clear, I don't see any problems with the code as it is now, but it still feels a bit risky

I had something like this in mind, but let me know what you think

```
# maybe not needed, but we can set this at the start to be safe
statuses = [status_pb2.Status(code=code_pb2.Code.UNKNOWN)] * len(batch)
try:
    operation = CrossSync._MutateRowsOperation(...)
    await operation.start()
except MutationsExceptionGroup as e:
    statuses = _create_statuses_from_exception(batch, e)
    # strip index information from exceptions, since it is not useful in a batch context
    for subexc in e.exceptions:
        subexc.index = None
    return list(e.exceptions)
else:
  statuses = [status_pb2.Status(code=code_pb2.Code.OK)] * len(batch)
finally:
    # mark batch as complete in flow control
    await self._flow_control.remove_from_flow(batch)

    # Call batch done callback with list of statuses.
    if self._user_batch_completed_callback:
        self._user_batch_completed_callback(statuses)
return []
```

google/cloud/bigtable/data/_helpers.py
```diff
@@ -224,6 +228,66 @@ def _align_timeouts(operation: float, attempt: float | None) -> tuple[float, flo
...
+    # 2. In the case of a singular exception, if the error does not have a gRPC status
+    #    code, we return a status code of UNKNOWN.
+    #
+    # 3. In the case of a RetryExceptionGroup, we use terminal exception in the exception
+    #    group and process that.
+    statuses = [status_pb2.Status(code=code_pb2.OK)] * batch_size
+    for error in exc_group.exceptions:
+        if isinstance(error.index, int) and 0 <= error.index < len(statuses):
+            cause = error.__cause__
+            if isinstance(cause, RetryExceptionGroup):
```
daniel-sanche, line 259: Do we need to do an extra check here to make sure cause.exceptions has items? I can't remember how that works.

If so, there could be potential for an IndexError here

google/cloud/bigtable/data/_helpers.py
```diff
@@ -224,6 +228,66 @@ def _align_timeouts(operation: float, attempt: float | None) -> tuple[float, flo
...
+                statuses[error.index] = _get_status(cause.exceptions[-1])
+            else:
+                statuses[error.index] = _get_status(cause)
+    return statuses
+
+
+def _get_status(exc: Exception) -> status_pb2.Status:
```
daniel-sanche: it looks like this can be called with None, if error.\_\_cause__ is None? Or is that not possible?

google/cloud/bigtable/data/_helpers.py
```diff
@@ -224,6 +228,66 @@ def _align_timeouts(operation: float, attempt: float | None) -> tuple[float, flo
...
+    ):
+        return status_pb2.Status(  # type: ignore[unreachable]
+            code=exc.grpc_status_code.value[0],
+            message=exc.message,
+            details=exc.details,
+        )
+
+    return status_pb2.Status(
+        code=code_pb2.Code.UNKNOWN,
+        message=str(exc),
```
daniel-sanche: If exc can be None, we should probably change this message?

google/cloud/bigtable/data/_helpers.py
gkevinzheng: I did some testing and Python prevents you from initializing an exception group without items, so we probably don't need to worry about the empty case.

daniel-sanche: Yeah, looks like that's consistent with our custom object: https://github.com/googleapis/python-bigtable/blob/c3ea54ee958a79959a649dce2553a84ec808b4d7/google/cloud/bigtable/data/exceptions.py#L81

daniel-sanche (approved):
LGTM

PR #1309, opened 2026-03-16
feat: Mutations Batcher shim

gkevinzheng:
**Changes Made:**

- Replaced mutations batcher implementation with one based off of the data client.
- Reworked unit tests.
- Added additional system tests.


google/cloud/bigtable/batcher.py
```diff
@@ -214,29 +87,41 @@ def __init__(
...
-        )
         self.table = table
-        self._executor = concurrent.futures.ThreadPoolExecutor()
-        atexit.register(self.close)
-        self._timer = threading.Timer(flush_interval, self.flush)
-        self._timer.start()
-        self.flow_control = _FlowControl(
-            max_mutations=MAX_OUTSTANDING_ELEMENTS,
-            max_mutation_bytes=MAX_OUTSTANDING_BYTES,
-        )
```
daniel-sanche, line 228: It looks like you're discarding the flow control? You should be able to [pass these through](https://github.com/googleapis/python-bigtable/blob/c3ea54ee958a79959a649dce2553a84ec808b4d7/google/cloud/bigtable/data/_sync_autogen/mutations_batcher.py#L189) to the data client batcher

google/cloud/bigtable/batcher.py
```diff
@@ -214,29 +87,41 @@ def __init__(
...
+
+    def _init_batcher(self):
+        self._batcher = self.table._table_impl.mutations_batcher(
+            flush_interval=self._flush_interval,
+            flush_limit_mutation_count=self._flush_count,
+            flush_limit_bytes=self._max_row_bytes,
+        )
+        self._batcher._user_batch_completed_callback = (
+            self._user_batch_completed_callback
+        )
```
daniel-sanche, line 115: It looks like you're only storing self._flush_interval, self._flush_count, and self._max_row_bytes so you have the context to re-build a new batcher later?

This pattern is fine, but I think functool.partial can make this kind of thing cleaner, letting you condensing all the state into single nullable function

```
self._batcher_build_fn = partial(self._build_batcher, callback, interval=flush_interval, ...)

def _build_batcher(self, callback, **kwargs):
    batcher = self.table._table_impl.mutations_batcher(**kwargs)
    batcher._user_batch_completed_callback = callback
    return batcher
```

And then every time you need a new batcher, you just call `self._batcher = self._batcher_build_fn()`

(It could be simplified more to get rid of the extra _build_batcher function if we make _user_batch_completed_callback into an unadvertised kwarg in the data client, but I'm not sure if that's worth it)

google/cloud/bigtable/data/_async/mutations_batcher.py
```diff
@@ -276,7 +276,9 @@ def __init__(
         self._newest_exceptions: deque[Exception] = deque(
             maxlen=self._exception_list_limit
         )
-        self._user_batch_completed_callback = None
+        self._user_batch_completed_callback: Optional[
```
daniel-sanche, line 280: we should probably add a comment here describing that this is currently just used by the shim

google/cloud/bigtable/batcher.py
```diff
@@ -406,9 +194,8 @@ def close(self):
         :raises:
             * :exc:`.batcherMutationsBatchError` if there's any error in the mutations.
```
daniel-sanche: I think the exc name in the docstrings might be broken?

google/cloud/bigtable/batcher.py
```diff
@@ -214,29 +87,41 @@ def __init__(
...
+        )
+
+    def _close_batcher(self):
+        try:
+            self._batcher.close()
+        except MutationsExceptionGroup as exc_group:
+            for error in exc_group.exceptions:
+                # Return the cause of the FailedMutationEntryError to the user,
+                # as this might be more what they're expecting.
+                self._exceptions.put(error.__cause__)
```
daniel-sanche, line 124: This might be a good time to address [this TODO](https://github.com/googleapis/python-bigtable/blob/4621e30a1849bf01021169387b2f3c58bf886e9a/google/cloud/bigtable/data/exceptions.py#L144)?

The code seems to assume it will be FailedMutationEntryError, so we should make the types agree

google/cloud/bigtable/batcher.py
```diff
@@ -214,29 +87,41 @@ def __init__(
...
+            self._user_batch_completed_callback
+        )
+
+    def _close_batcher(self):
+        try:
+            self._batcher.close()
+        except MutationsExceptionGroup as exc_group:
+            for error in exc_group.exceptions:
+                # Return the cause of the FailedMutationEntryError to the user,
+                # as this might be more what they're expecting.
```
daniel-sanche: Can this comment be improved, to be more definitive? Maybe something like `# Unpack the root cause from FailedMutationEntryError wrapper`

daniel-sanche (changes requested):

google/cloud/bigtable/data/_async/mutations_batcher.py
```diff
@@ -519,26 +524,27 @@ def _on_exit(self):
...
+            CrossSync.Future[list[FailedMutationEntryError]] | CrossSync.Future[None]
+        ],
+        batches: Sequence[list[RowMutationEntry]],
+    ) -> list[FailedMutationEntryError]:
         """
         Takes in a list of futures representing _execute_mutate_rows tasks,
         waits for them to complete, and returns a list of errors encountered.
 
         Args:
             *tasks: futures representing _execute_mutate_rows or _flush_internal tasks
```
daniel-sanche: this Args needs to be updated

google/cloud/bigtable/data/_async/mutations_batcher.py
```diff
@@ -519,26 +524,27 @@ def _on_exit(self):
...
+                list of FailedMutationEntryError encountered by any of the tasks,
+                representing a failed mutation operation.
+                Successful tasks will not be represented in the output list.
         """
         if not tasks:
             return []
-        exceptions: list[Exception] = []
-        for task in tasks:
+        exceptions: list[FailedMutationEntryError] = []
+        for task, batch in list(zip(tasks, batches)):
```
daniel-sanche: This feels risky, since zip will stop early if one list is a different size than the other. This could turn into a data loss issue in the future

Maybe we should have a single list of tuples, so we have a bit more type safety there?

google/cloud/bigtable/data/exceptions.py
daniel-sanche: remove this TODO. And update the docstring: 

> Exceptions will typically be of type FailedMutationEntryError, but other exceptions may
> be included if they are raised during the mutation operation

daniel-sanche (changes requested):
We need to be careful with that zip call, and there are a couple docstrings to update. Otherwise looks good

daniel-sanche (approved):
LGTM

